PR #121, opened 2020-03-04
Piping Support

### LINKED ISSUE #90: Investigate how to map key names when pipelining input object.s
Instead of having to do this:

`get-GroupMember -GroupId $groupid | % { write-output @{ UserId=$_.id } } | get-user`

we should be able to do this,

`get-GroupMember -GroupId $groupid | get-user`

peombwa:
This PR adds support for piping of cmdlets by applying an AutoREST transform directive on the generated cmdlets.

### Problem
`Get-MgUser | Get-MgUserPresence` fails with error InputObject has null value for InputObject.UserId.

When piping two cmdlets, the cmdlet receiving the response from the first cmdlet attempts to retrieve an entity's id property as `entity-id`, yet the response object has it as `Id`. This is because our OpenAPI docs specifies path parameters as `entity-id` e.g. `user-id`.
image
The generated `InputObject` - `UsersUserIdentity` -  ends up looking for `entity-id` property which is not present.
image

### Solution in this PR
The directives adds the following : 
- The first directive adds partial `AfterToJson(JsonObject container, SerializationMode serializationMode);` signature overload to all `/generated/api/Models/MicrosoftGraph*.json.cs`. `AfterToJson` will be called after json serialization, allowing for customization of the `JsonObject` before it's deserialized to an `InputObject` which is used during piping.
image
The directive then adds a call to `AfterToJson(ref container, serializationMode);` on `ToJson` method. `ToJson` serializes an instance of a model e.g. `MicrosoftGraphUser`.
image

- The second directive add an implementation `AfterToJson(JsonObject container, SerializationMode serializationMode)` to all models (`/generated/api/Models/MicrosoftGraph*.cs`) that have an `Id` property except for `MicrosoftGraphEntity` and `MicrosoftGraphBaseItem`. The implementation adds an `entity-id` e.g. `user-id` to the `JsonObject` being serialized.
image

### Sample
image

Closes #90


darrelmiller:
I had not realized that it was the OpenAPI that was causing us to generate a parameter name that does not match the object property.  Initially the OpenAPI parameters were all using the name "Id" and we had to change that so that paths with two parameters would have unique names.  We could change the parameter naming so that the last Id in the path keeps the "Id" name.  I believe that would solve most of the problems.   However, the solution you have created has the added benefit that piping in an object that contains multiple path parameters would actually match.
Let's go with this approach and do some testing. At least this gives us a backup plan.

broo0ose:
This is a great improvement since I last looked at it. 
I can't do: 
`Get-MgUser | Get-MgUserPresence`

But this works:
`Get-MgUser -userID "someone@somewhere.com" |  %{Get-MgUserPresence -UserID $_.Id}`

Issue #154, opened 2020-03-24
Connect-Graph fails to cast roles into String[] if there is only 1 permission


genorex:
An example below is when there is only 1 API application permission:
Connect-Graph : Error converting value "Reports.Read.All" to type 'System.String[]'. Path 'roles', line 1, position 441.

darrelmiller:
@genorex  Can you provide some details of the module versions that you are using?  We are not able to repro this.

peombwa:
@genorex This issue is due to how we are reading the claims from the access token. This will be fixed by PR https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/164.

Issue #124, opened 2020-03-06
Connect-Graph : Authentication needed when call is within a function


daveschafer:
When I use the Graph Module inline, everything works as expected, for example:

```powershell
Connect-Graph -ClientId $ClientId -TenantId $TenantId -CertificateThumbprint $CertificateThumbprint 
Get-MgUser -Top 1
```
This provides the expected Output.  

Although when I use Connect-Graph within a Function, it fails:

```powershell
function test-graph {
    $ClientId = "<omitted>"
    $TenantId ="<omitted>"
    $CertificateThumbprint ="<omitted>"
    
    Connect-Graph -ClientId $ClientId -TenantId $TenantId -CertificateThumbprint $CertificateThumbprint -ForceRefresh
    Get-MgUser -Top 1
}

test-graph
```
This fails with the Message:
```
Welcome To Microsoft Graph!
Get-MgUser_List: 
Line |
   8 |      Get-MgUser -Top 1
     |      ~~~~~~~~~~~~~~~~~
     | Authentication needed, call Connect-Graph.
```

I can reproduce this Error in PS 5.1 and in PS7. I also tried different ways of installing and importing the Graph Module (Global), without improvement.

bodangly:
I can add to this and say I have the same behavior. 

Some things to note. If I don't give it a -TenantId, I will get the prompt to do a devicelogin, even if called within a function. However, this context isn't persisted and all further mg functions fail. I have to login from an interactive prompt as stated for the context to be preserved.

If I do specify a -TenantId, I don't even get the devicelogin prompt, it just outputs "Welcome to Microsoft Graph!" (if I pipe to Write-Host in the script), and no context actually gets captured.

Issue #113, opened 2020-02-18
Non-interactive authentication does not work when used from a PowerShell script


AnatoliB:
The following script fails:

``` PowerShell
Import-Module Microsoft.Graph.Authentication
Import-Module Microsoft.Graph.Users.User

Connect-Graph -TenantId [REDACTED] -ClientId [REDACTED] -CertificateName [REDACTED] -CertificateThumbprint [REDACTED]

Get-MgUser
```
`Connect-Graph` succeeds, but `Get-MgUser` reports the following error: **Authentication needed, call Connect-Graph.**

Exactly the same commands executed interactively on a PowerShell console run successfully, so this does not look like a configuration/environment issue. In order to repro:

1. Save this code into a test.ps1 file.
2. **IMPORTANT**: In a brand new PowerShell session, invoke test.ps1 like this: `.\test.ps1` or this: `& .\test.ps1`. (If you run the `Connect-Graph` and `Get-MgUser` commands by invoking these commands one by one in a PowerShell console, or dot-sourcing the same script (`. .\test.ps1`), these commands succeed and the issue does **not** repro.)

As a workaround, inserting the following line immediately after Connect-Graph invocation fixes the problem:
``` PowerShell
$global:GraphAuthConfigId = $GraphAuthConfigId
```

While debugging this, we noticed that `Get-MgUser` cannot find the `GraphAuthConfigId` variable, even though it is definitely initialized by `Connect-Graph`. We can even print it out from the script scope, and the content looks correct. I suspect the variable is created in a way that makes it invisible in the `Get-MgUser` function scope, and this is why copying it into the global scope helps. Perhaps this should be fixed in the code that creates the variable.

AnatoliB:
The variable is initialized at: https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/e64d120cbb53db7f1b80a73c8b80e66de0814b8b/src/Authentication/Authentication/Cmdlets/ConnectGraph.cs#L76

Perhaps a `PSVariable` object should be created with a constructor that takes `ScopedItemOptions`, passing something like `ScopedItemOptions.AllScope`. I haven't tried this fix, just an idea.

darrelmiller:
Hey @peombwa could you take a look at this, it is block usage in Azure Functions.  Thanks.

AnatoliB:
Please let us know if you have any ETA on this.

PR #164, opened 2020-04-08
Expose AuthContext To Function Scope

peombwa:
This PR adds a `GraphSession` class which safely provides a single session instance of itself across multiple Microsoft Graph modules. This session object will be use to hold the `AuthContext`, thus making it accessible to both Powershell scripts/functions and inline calls. This closes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/124 and https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/113.

The PR also fixes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/154, by changing how the JWT token is decoded.


ddyett:
>             SessionState.PSVariable.Remove(Constants.GraphAuthConfigId);

should this not clear the graphsession state?

---
Refers to: src/Authentication/Authentication/Cmdlets/DisconnectGraph.cs:34 in b0c31bc. [](commit_id = b0c31bc89faca6e6e5f19c3a376dfaf040438f28, deletion_comment = False)

src/Authentication/Authentication/Models/AuthContext.cs
```diff
@@ -1,16 +1,12 @@
...
+namespace Microsoft.Graph.PowerShell.Authentication
 {
-    public enum AuthenticationType
-    {
-        Delegated,
-        AppOnly
-    }
-
-    public class AuthConfig
+    using Newtonsoft.Json;
```
ddyett, line 6: >using Newtonsoft.Json; [](start = 4, length = 22)

since we're doing the work to convert these in the core library do we have an issue for updating these?

src/Authentication/Authentication/Common/GraphSession.cs
```diff
@@ -0,0 +1,128 @@
...
+                {
+                    throw new InvalidOperationException(ErrorConstants.Codes.SessionLockReadRecursion, lockException);
+                }
+                catch (ObjectDisposedException disposedException)
+                {
+                    throw new InvalidOperationException(ErrorConstants.Codes.SessionLockReadDisposed, disposedException);
+                }
+            }
+        }
+        public GraphSession()
```
ddyett, line 53: NIT: spacing

src/Authentication/Authentication/Common/GraphSession.cs
```diff
@@ -0,0 +1,128 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph.PowerShell.Authentication
+{
+    using System;
+    using System.Threading;
+    /// <summary>
+    /// The current <see cref="GraphSession"/>.
```
ddyett, line 10: >The current <see cref="GraphSession"/> [](start = 8, length = 38)

you can probably add a few more words about what the class is meant to do.

src/Authentication/Authentication/Common/GraphSession.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+                catch (ObjectDisposedException disposedException)
+                {
+                    throw new InvalidOperationException(ErrorConstants.Codes.SessionLockReadDisposed, disposedException);
+                }
+            }
+        }
+        public GraphSession()
+        {
+            _graphSessionId = Guid.NewGuid();
```
ddyett, line 55: > _graphSessionId = Guid.NewGuid(); [](start = 11, length = 34)

why not do this when you initialize and take away the ability to directly construct a graphsession all together.

ddyett: ok it's because of how you want to possibly initialize from a file in the future.

---
In reply to: [405665227](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/164#discussion_r405665227) [](ancestors = 405665227)

src/Authentication/Authentication.Test/Helpers/GraphSessionTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void GraphSessionShouldBeInitilizedAfterInitializerIsCalled()
+        {
+            GraphSessionInitializer.InitializeSession();
+
+            Assert.NotNull(GraphSession.Instance);
+            Assert.Null(GraphSession.Instance.AuthContext);
+        }
+        
+        [Fact]
+        public void ShoudlOverwriteExistingGraphSession()
```
ddyett, line 18: >Shoudl [](start = 20, length = 6)

Should is misspelt

src/Authentication/Authentication.Test/Helpers/GraphSessionTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            Guid originalSessionId = GraphSession.Instance._graphSessionId;
+
+            GraphSession.Initialize(() => new GraphSession(), true);
+
+            Assert.NotNull(GraphSession.Instance);
+            Assert.NotEqual(GraphSession.Instance._graphSessionId, originalSessionId);
+        } 
+        
+        [Fact]
+        public void ShoudlNotOverwriteExistingGraphSession()
```
ddyett, line 30: >Shoudl [](start = 20, length = 6)

misspelt.

src/Authentication/Authentication/Common/GraphSession.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="instanceCreator">A func to create an instance.</param>
+        /// <param name="overwrite">If true, overwrite the current instance. Otherwise do not initialize.</param>
+        public static void Initialize(Func<GraphSession> instanceCreator, bool overwrite)
+        {
+            try
+            {
+                sessionLock.EnterWriteLock();
+                try
+                {
+                    if (overwrite || !_initialized)
```
ddyett, line 70: >overwrite [](start = 24, length = 9)

the pattern here if you don't set the flag but has an instance seems like silent failure which leaves code error prone.  It should probably throw an exception that it is not overwriting the session.

src/Authentication/Authentication/Common/GraphSession.cs
peombwa: That's right. We will use the initializer to read the auth context from a file and even support multiple login accounts.

src/Authentication/Authentication/Models/AuthContext.cs
peombwa: The PS SDK still uses the old core library. We will only need to serialize this class when we start storing auth context locally. I'll remove the attribute for now.

src/Authentication/Authentication/Common/GraphSession.cs
peombwa: The assumption made here is that `Initialize` should only be called once, and if one wants to overwrite the current instance, then they have to do so by explicitly calling its overload that accepts `bool overwrite`.

ddyett: your intent is right.  The issue is that if you do call twice you won't know that it did not give you back a new GraphSession and making that assumption can lead to further bugs in the callers code.  That's why I think it's better to throw and say you must explicitly overwrite so that assumption that it did modify cannot be made.

PR #176, opened 2020-04-17
Set root as default for SiteId

finsharp:
Set root as the default for SiteId in Get-MgSite


finsharp:
Defaulting to "root" for

- Get-MgSite
- Get-MgSiteAnalytic
- Get-MgSiteColumn
- Get-MgSiteContentType
- Get-MgSiteContentTypeContentLink
- Get-MgSubSite

Defaulting to root for Update-*, New-* and Remove-* would be a risky default especially for users with admin privileges.

darrelmiller:
If we have a default value for this, how do we differentiate between getting a list of sites and getting a root site   /cc @peombwa @finsharp

peombwa:
@darrelmiller You can't get a list of sites using the SharePoint APIs. You can only get a list of subsites which requires a siteId. The only thing that's supported on `/sites` is search which can be called by  `-search` parameter.

darrelmiller:
@peombwa  You actually can get a list of sites if you pass a special header.  They are bringing that to be part of the public API for Build.

peombwa:
I see, then we will have to exclude `Get-MgSite`, or add `Get-MgSiteRoot` cmdlet cc @finsharp.

Issue #11, opened 2019-07-08
Investigate paging of collections


darrelmiller:
[AB#4595](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4595)

peombwa:
Investigate how to support paging of collections using NextLink.

ChrisB9907:
There is currently no way to get more than 999 users without paging and the -skip param isn't available on Get-MgUser

darrelmiller:
@ChrisB9907 That is correct.  We have prototyped a solution that enables automatic paging and once we figure out how to turn it off by default, we will add it to our packages.

Windos:
My preference is adding a -All switch parameter that returns all.

The key to avoid backlash, regardless of the name that ends up being used is **not** making it a Boolean parameter.

I [really](https://twitter.com/WindosNZ/status/1016791902841028609?s=20) [hate](https://twitter.com/WindosNZ/status/1104863709954891776?s=20) AzureAD's [implementation](https://twitter.com/WindosNZ/status/1220231277644677120?s=20) with ["-All $true"](https://twitter.com/WindosNZ/status/1236782220691398657)

darrelmiller:
## Option 1

Get-MgUser   -> Return first page
Get-MgUser -top 100    -> return a page of 100
Get-MgUser -top All    -> would page all results
Get-MgUser -top 2000    -> would page all results up to 2000
Get-MgUser -Limit 2000    -> Limit is an alias top

Can we make paging off by default?
Can we bypass the top paramter integer datatype to add support for all?
Can we implement our on limit on a set of pages.

## Option 2

Get-MgUser   -> Return first page
Get-MgUser -top 100    -> return a page of 100
Get-MgUser -All   -> return all pages
Get-MgUser -All -top 2000   -> return all pages up to 2000
Get-MgUser -top 2000   -> Values larger than 999 cause implicit -All
Get-MgUser -Limit 2000   -> Limit is an alias to top

ChrisB9907:
[...]

PR #185, opened 2020-04-30
Modules Refresh

peombwa:
This PR:
- Refreshes OpenAPI docs with last weeks (04/20/2020) MS graph CSDL. The current CSDL has a wrongly named `request` entity that has to be renamed.
- Adds `Identity.AuthenticationMethods` module - [doc page](https://docs.microsoft.com/en-us/graph/api/resources/authenticationmethods-overview?view=graph-rest-beta).
- Temporarily ignores handling of `ODataNextLink` by cmdlets since this is WIP that will be addressed by #11 . 
image


ddyett:
Do we need the identity.authenticationmethods?  It looks like a duplicate of the users.Authentication

peombwa:
> Do we need the identity.authenticationmethods? It looks like a duplicate of the users.Authentication

Great catch! It sneaked in as part of a module name change that I made. I initially had it as `Users.Authentication`, then renamed it to `Identity.AuthenticationMethods` so that it aligns with our [doc page ToC](https://docs.microsoft.com/en-us/graph/api/resources/authenticationmethods-overview?view=graph-rest-beta) and its workload owner (Identity).

PR #203, opened 2020-05-05
Download and Create with OpenApiDocs every week

finsharp:
Download OpenApiDocs from GraphExplorerApi and Create a PR.

Generated PR [OpenApiDocs PR](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/202)


.azure-pipelines/download-openapidocs.yml
```diff
@@ -0,0 +1,112 @@
...
+      displayName: "Weekly OpenApiDocs Download and PR"
+      branches: 
+        include:
+        - dev
+      always: true
+
+steps:
+  - checkout: self
+    persistCredentials: true
+    clean: true
```
MIchaelMainer, line 24: nit:

`fetchDepth: 1`, no need to download the entire history.

.azure-pipelines/download-openapidocs.yml
```diff
@@ -0,0 +1,112 @@
...
+      targetType: inline
+      script: |
+        $branch = "{0}.{1}" -f "weeklyOpenApiDocsDownload", (Get-Date -Format yyyyMMdd)
+        Write-Host "##vso[task.setvariable variable=BRANCH;]$branch"
+  - task: PowerShell@2
+    displayName: "Configure User"
+    inputs:
+      targetType: 'inline'
+      script: |
+        git config --global user.email "mgpssdk@microsoft.com"
```
MIchaelMainer, line 42: Why are you using this email?

.azure-pipelines/download-openapidocs.yml
```diff
@@ -0,0 +1,112 @@
...
+name: $(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+
+pool:
+  vmImage: "windows-latest"
+
+variables:
+  BRANCH: 'weeklyOpenApiDocsDownload'
+
+schedules:
+    - cron: "0 12 * * 5" # Run Every Friday at Midnight UTC Time
```
peombwa, line 14: For it to trigger at Midnight UTC, the [cron](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#supported-cron-syntax) has to be `0 0 * * 5`. 

@MIchaelMainer At what time does the metadata get updated? We would ideally want to sync this after the metadata has been update. Maybe we can run it a few hours after the CSDL captured? @ddyett, @darrelmiller, thoughts?

MIchaelMainer: Tuesday morning capture. If this is run on Wednesday, it gives us time to unblock ourselves via transform.

darrelmiller: We should also investigate if we can trigger on a commit to the onboarding repo.  The problem with using the timer based trigger is when the AGS team do a rollback we miss it.

MIchaelMainer: Does the DevX API request a new CSDL file from AGS when we make a request for OpenAPI spec? Where and when does the DevX API get the CSDL? 

To be clear, I have no idea how the CSDL gets to production. I'd like to know how that happens.

finsharp: @darrelmiller [web hooks](https://docs.microsoft.com/en-us/azure/devops/service-hooks/services/webhooks?view=azure-devops) might be a viable solution.

finsharp: Due to changes and accommodations in the DevX API, having a Wednesday schedule seems satisfactory to get the latest OpenApiDocs.

peombwa:
@finsharp It looks like the ADO pipeline job is failing with a 401 when creating a PR https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=18084&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=dee4e6ec-be1b-5671-4678-ab293b4004c5&l=83.

finsharp:
Using GITHUB_TOKEN as a variable instead of System.AccessToken. 
@peombwa Please take another look.

finsharp:
@MIchaelMainer We are using a PAT from my account for this, is there a way to use the Service Connection token to do this?

MIchaelMainer:
No PAT please. 

I would have thought that $(System.AccessToken) would work for the Azure Pipeline connection. [Then I learn](https://stackoverflow.com/questions/56735809/what-is-the-scope-of-the-token-in-system-accesstoken-in-azure-devops-pipeline): System.Accesstoken is a PAT token generated for the service identity “Project Collection Build Service (account)”

finsharp:
Considering @MIchaelMainer observations, should we move this back to the initial Github Action work/structure?

MIchaelMainer:
To be clear, I mean a GitHub action just for creating a PR based on a branch pattern match.

finsharp:
@MIchaelMainer I have removed the PR creation process from the Azure Pipeline, and move PR creation to Github Actions. 
When a push to 'weeklyOpenApiDocsDownload*' is detected, the PR is created.

finsharp:
This [PR](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/235) was automatically created by the Github action after Azure Pipelines downloaded and committed the files.

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  push:
+    branches: 
+      - 'weeklyOpenApiDocsDownload.*'
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  downloadOpenApiDocs:
+    # The type of runner that the job will run on
```
MIchaelMainer, line 17: I suggest you add a skip ci option.
https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/blob/master/.github/workflows/create-pull-request.yml#L24

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# This is a basic workflow to help you get started with Actions
+
+name: WeeklyOpenApiDocsDownload
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches:
```
MIchaelMainer, line 10: Is there a possibility that you need a path spec restriction on creating PRs? Probably not because of the specific branch spec.

.github/workflows/main.yml
MIchaelMainer: This is particularly useful if you need to update the PR with a commit, so as not to trigger this GH action again.

Issue #63, opened 2019-12-03
Encrypt tokens!


michevnew:
Any particular reason why you are storing the tokens in unencrypted form on the hard drive? Hardly the best security practice.

I understand that you might be using this for ensuring session longevity and such, but at least store them as secure string, or use the cred man or something. On a related note, why arent you leveraging the ADAL/MSAL token cache?

[AB#4594](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4594)

darrelmiller:
Thanks for calling this out.  We had tokens being encrypted on Windows, but we didn't have it working on Linux/OSX because there was no support for it in MSAL.  @peombwa worked on adding support for encrypting tokens on Linux last week and we expect to be able to add encryption back very soon.

We are using the MSAL Token cache.  We access it here https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs#L48

michevnew:
I actually noticed this on a Windows machine. The userTokenCache.bin3 is created in the assembly directory and contains the unencrypted tokens, you can simply copy/paste it for use in auth header outside of the Graph module. Using 0.1.4 if it matters.

As long as you are aware of this an plan to address it, I'm fine :)

darrelmiller:
I'm not sure why we needed to remove it temporarily for Windows, but I do know that this is high on our priority to fix.

peombwa:
We had to temporarily remove it for Windows because System.Security.Cryptography.ProtectedData doesn't support PowerShell 5 (.Net 4.5). This is something high on our priority list.

PR #216, opened 2020-05-08
Multi-Platform Token Cache

peombwa:
This PR adds support for a multi-platform token cache since PowerShell core runs on Linux, MacOS and Linux. This is implemented as follows:
#### Windows
 On Windows, the token is encrypted using [Windows DPAPI](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.protecteddata?view=netframework-4.8#remarks) and stored locally on a file. Each app/client id will have its own token cache file.

#### MacOS
The token is stored in the users default keychain - login keychain. This is done by calling native keychain APIs exposed via [Security framework](https://developer.apple.com/documentation/Security).
image

#### Linux
On Linux, the token is stored in the [kernel's key management facility](https://www.kernel.org/doc/html/v4.13/security/keys/core.html?highlight=keyctl#userspace-system-call-interface) via [keyutils](http://man7.org/linux/man-pages/man7/keyutils.7.html), and linked to the [session keyring.](http://man7.org/linux/man-pages/man7/session-keyring.7.html)
image

#### Windows
[Test logs](https://o365exchange.visualstudio.com/O365%20Sandbox/_build/results?buildId=2896644&view=logs&j=d03e2b71-7237-5b58-51fa-970c18c176d4).
#### Linux
image
[Test logs.](https://o365exchange.visualstudio.com/O365%20Sandbox/_build/results?buildId=2896644&view=logs&j=1205dcaf-72d3-5264-1d73-e610530a0ff7)

#### MacOS
image
[Test logs.](https://o365exchange.visualstudio.com/O365%20Sandbox/_build/results?buildId=2896644&view=logs&j=26f09245-25f3-57fa-164a-3d3a2cbb184b)

#### Linux
- [GNOME keyring](https://wiki.gnome.org/Projects/GnomeKeyring/) was considered as a replacement for the native P/Invoke calls made to `keyutils`. Unfortunately, the package is not available on [Azure Cloud Shell](https://azure.microsoft.com/en-us/features/cloud-shell/), and most Linux distros.
- KeyUtils doesn't persist its data after restarts.

#### MacOS
- [Xamarin.Essentials](https://docs.microsoft.com/en-us/xamarin/essentials/) was considered as a replacement for the native P/Invoke calls made to `Security Foundation`. Unfortunately, Xamarin.Essentials doesn't support MacOS at this time.

Closes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/63 [AB#4716](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4716)


.azure-pipelines/validate-pr-auth-module.yml
```diff
@@ -49,4 +49,28 @@ jobs:
...
+  displayName: MS Graph PS SDK Auth Validation - MacOS
+  pool:
+    vmImage: 'macOS-latest'
+  steps:
+  - task: DotNetCoreCLI@2
+    displayName: 'Run Enabled Tests'
+    inputs:
+      command: 'test'
+      projects: '$(System.DefaultWorkingDirectory)/src/Authentication/Authentication.Test/*.csproj'
+      testRunTitle: 'Run Enabled Tests'
```
ddyett, line 76: shouldn't these also have a fail notification

src/Authentication/Authentication/TokenCache/TokenCacheStorage.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// Helper class to handle token encryption and decryption.
+    /// </summary>
+    internal static class TokenCacheStorage
+    {
+        /// <summary>
+        /// Gets a decrypted access token from the host platform OS.
+        /// </summary>
+        /// <param name="appId">An app/client id.</param>
+        /// <returns></returns>
+        public static byte[] GetToken(string appId)
```
ddyett, line 16: >appId [](start = 45, length = 5)

input validation

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -46,37 +46,24 @@ internal static void Logout(IAuthContext authConfig)
...
-            string tokenCacheFilePath = Path.Combine(Constants.TokenCacheDirectory, tokenCacheFile);
-
             tokenCache.SetBeforeAccess((TokenCacheNotificationArgs args) => {
                 lock (FileLock)
                 {
-                    args.TokenCache.DeserializeMsalV3(File.Exists(tokenCacheFilePath)
-                        ? TokenCryptographer.DecryptToken(File.ReadAllBytes(tokenCacheFilePath))
-                        : null,
-                        shouldClearExistingCache: true);
+                    args.TokenCache.DeserializeMsalV3(TokenCacheStorage.GetToken(appId), shouldClearExistingCache: true);
```
ddyett, line 58: >TokenCacheStorage.GetToken(appId) [](start = 54, length = 33)

is it safe to call Deserialize if this returns nothing?

.azure-pipelines/validate-pr-auth-module.yml
```diff
@@ -14,8 +14,8 @@ pr:
 trigger: none
 
 jobs:
-- job: MSGraphPSSDKValidation
-  displayName: MS Graph PS SDK Auth Validation
+- job: MSGraphPSSDKValidation_Windows
+  displayName: MS Graph PS SDK Auth Validation - Windows
```
MIchaelMainer, line 18: We should move this off of the Microsoft Graph pool. We should only use this pool for pipeline debugging/development.

src/Authentication/Authentication/Constants.cs
```diff
@@ -13,8 +13,7 @@ public static class Constants
         internal const string UserParameterSet = "UserParameterSet";
         internal const string AppParameterSet = "AppParameterSet";
         internal const int MaxDeviceCodeTimeOut = 120; // 2 mins timeout.
-        internal const string UserCacheFileName = "userTokenCache.bin3";
-        internal const string AppCacheFileName = "appTokenCache.bin3";
         internal static readonly string TokenCacheDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".graph");
+        internal const string TokenCahceServiceName = "com.microsoft.graph.powershell.sdkcache";
```
MIchaelMainer, line 17: typo: TokenCahceServiceName -> TokenCacheServiceName

src/Authentication/Authentication/TokenCache/LinuxTokenCache.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Gets an app's token from Linux kerings faciility.
+        /// </summary>
+        /// <param name="appId">An app/client id.</param>
+        /// <returns>A decypted token.</returns>
+        public static byte[] GetToken(string appId)
+        {
+            int key = LinuxNativeKeyUtils.request_key(
+                type: LinuxNativeKeyUtils.KeyTypes.User,
+                description: $"{Constants.TokenCahceServiceName}:{appId}",
```
MIchaelMainer, line 25: typo TokenCahceServiceName - > TokenCahceServiceName

src/Authentication/Authentication/TokenCache/LinuxTokenCache.cs
```diff
@@ -0,0 +1,94 @@
...
+    using System;
+    using System.Runtime.InteropServices;
+
+    /// <summary>
+    /// A set of methods for getting, setting and deleting tokens on Linux via keyutils.
+    /// </summary>
+    internal static class LinuxTokenCache
+    {
+        /// <summary>
+        /// Gets an app's token from Linux kerings faciility.
```
MIchaelMainer, line 17: typo: kerings faciility -> keyrings facility

src/Authentication/Authentication/TokenCache/TokenCacheStorage.cs
MIchaelMainer: :point_up: this is a global comment

src/Authentication/Authentication.Test/TokenCache/TokenCacheStorageTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                    failed = true;
+                }
+
+                CleanTokenCache(index.ToString());
+                count++;
+            });
+
+            // Assert
+            Assert.Equal(executions, count);
+            Assert.False(failed);
```
zengin, line 140: consider having a customized message for these failures, it can help parse what these values represent without looking at the code.

src/Authentication/Authentication/TokenCache/LinuxTokenCache.cs
MIchaelMainer: 👆 global comment.

src/Authentication/Authentication/TokenCache/LinuxTokenCache.cs
MIchaelMainer: :point_up: global comment

src/Authentication/Authentication/TokenCache/WindowsTokenCache.cs
```diff
@@ -4,27 +4,62 @@
...
         /// </summary>
-        /// <param name="buffer">A <see cref="byte[]"/> to encrypt.</param>
-        /// <returns>An encrypted <see cref="byte[]"/>.</returns>
-        public static byte[] EncryptToken(byte[] buffer)
+        /// <param name="appId">The app/client id..</param>
+        /// <returns>A decrypted access token.</returns>
+        public static byte[] GetToken(string appId)
         {
-            return ProtectedData.Protect(buffer, null, DataProtectionScope.CurrentUser);
+            if (!Directory.Exists(Constants.TokenCacheDirectory))
```
zengin, line 21: >if (!Directory.Exists(Constants.TokenCacheDirectory)) [](start = 12, length = 53)

doesn't look like you need this check: [https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=netcore-3.1)

src/Authentication/Authentication/TokenCache/MacTokenCache.cs
```diff
@@ -0,0 +1,214 @@
...
+                    accountNameLength: (uint)appId.Length,
+                    accountName: appId,
+                    passwordLength: out uint passwordLength,
+                    passwordData: out passwordDataPtr,
+                    itemRef: out itemPtr);
+
+                if (resultStatus == MacNativeKeyChain.SecResultCodes.errSecItemNotFound)
+                    return contentBuffer;
+
+                else if (resultStatus != MacNativeKeyChain.SecResultCodes.errSecSuccess)
```
zengin, line 43: Opinion: I am not familiar with the current design guidelines of this repository, but it is hard to connect this `if` with the `else` block below without the curly brackets. To prevent accidental modifications, I personally prefer having the curly brackets even if there is a single statement in the block. If you have a reasoning against it, I would be happy to hear it.

src/Authentication/Authentication/TokenCache/MacTokenCache.cs
```diff
@@ -0,0 +1,214 @@
...
+                        ErrorConstants.Message.MacKeyChainFailed,
+                        "SecKeychainFindGenericPassword",
+                        resultStatus));
+                }
+
+                if (itemPtr != IntPtr.Zero && passwordLength > 0)
+                {
+                    contentBuffer = new byte[passwordLength];
+                    Marshal.Copy(passwordDataPtr, contentBuffer, 0, contentBuffer.Length);
+                }
```
zengin, line 56: NIT: this block can be a `switch` on the result code enum.

src/Authentication/Authentication/TokenCache/MacTokenCache.cs
```diff
@@ -0,0 +1,214 @@
...
+                    resultStatus != MacNativeKeyChain.SecResultCodes.errSecItemNotFound)
+                {
+                    throw new Exception(string.Format(
+                        CultureInfo.CurrentCulture,
+                        ErrorConstants.Message.MacKeyChainFailed,
+                        "SecKeychainFindGenericPassword",
+                        resultStatus));
+                }
+
+                if (itemPtr != IntPtr.Zero)
```
zengin, line 96: >itemPtr != IntPtr.Zero [](start = 20, length = 22)

Q: is this equivalent to resultStatus == MacNativeKeyChain.SecResultCodes.errSecSuccess? If so, it can be a good idea to change the result evaluation into a switch where the cases are self documenting: i.e. errSecItemNotFound will document the "// Key not found" message below.

src/Authentication/Authentication/TokenCache/MacTokenCache.cs
zengin: Even if you decide to keep it this way, you can remove `else` keyword, as the only statement above is a return.

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
peombwa: It's safe to return pass a null to `DeserializeMsalV3`. [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1cef43d185bb0a1dcc013d2ec232ae9d4c6c053a/src/client/Microsoft.Identity.Client/TokenCache.Serialization.cs#L42) does the check.

src/Authentication/Authentication/TokenCache/MacTokenCache.cs
peombwa: You are right. `itemPtr` is a pointer to a reference to the new keychain item. Thus, if a keychain item is found, `itemPtr` will point to a reference; if not, it will be null. I'll simplify this evaluation as per your suggestion.

peombwa:
> I am wondering what type of native API exceptions we need handle.

We are already handling native API exceptions on MacOS. We are using native exceptions - `errSecSuccess` and `errSecItemNotFound` - to determine code flow. All other exceptions are mapped to a generic .NET exception with their native result code as the message.

On Linux, the response from the native APIs is either 0 for success or -1 for a failure. If a failure occurs, we return an empty byte[] to MSAL. This defaults to the current behavior on Linux.

> Also, would it make sense to pull this out of the PowerShell SDK and make this a separate package that can be consumed by .Net SDK and really, anyone else that may to support .NET core on all 3 platforms?

MSAL team is already working on an extension to do just that. On Linux, they will be using using Gnome Keyrings which is not available on Azure Cloud Shell.

PR #232, opened 2020-05-18
Log HTTP Request and Response when Debug Switch is specified.

finsharp:
Hook into Signal infrastructure to capture HTTP Request and Responses for logging purposes when Debug is specified.

On Module init, hook EventHandler delegate into EventListener delegate.
Use HttpFormatter from contrib repo. Does not show the Authorization header (Access Token).

Depends on changes in autorest.powershell [generator](https://github.com/Azure/autorest.powershell/pull/622)

Covers task [AB#4720](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4720)


finsharp:
Changes to AutoRest.Powershell have been reviewed and merged.

tools/Custom/EventHelper.cs
```diff
@@ -0,0 +1,93 @@
+using System;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Graph.PowerShell.Runtime;
+
+namespace Microsoft.Graph.PowerShell
```
darrelmiller, line 7: I would make this an EventFactory class and move Print into an EventExtensions class.  The Helper suffix just isn't very helpful.

tools/Custom/EventHelper.cs
```diff
@@ -0,0 +1,93 @@
+using System;
```
darrelmiller, line 1: We should probably start putting the Microsoft copyright notice on files.

tools/Custom/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,356 @@
...
+        // Set of header fields that should get serialized as space-separated values such as User-Agent.
+        private static readonly HashSet<string> _spaceSeparatedValueHeaderFields = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "User-Agent",
+        };
+
+        // Set of header fields that should not get serialized
+        private static readonly HashSet<string> _neverSerializedHeaderFields = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "SdkVersion",
```
darrelmiller, line 49: Why are we not Serializing SdkVersion and FeatureFlag? Or Cache-Control, or Date?

finsharp: This is legacy from the contrib project.

finsharp:
Generated a sample module using the latest autorest.powershell that includes our code generator changes, generates correctly and `-Debug` emits Request and Responses accordingly.

tools/Custom/HttpMessageFormatter.cs
peombwa: We should serialize headers that are useful in tracing an issue/bug. For instance, we need SdkVersion to know which version of the SDK the customer was running, featureFlag to know which middleware handlers they have enabled e.t.c. `Authorization` header is not useful, and we shouldn't serialize it.

PR #241, opened 2020-05-26
Remove super-permissions from sample

psignoret:


darrelmiller:
@psignoret Thanks for this!  I will merge this in.

PR #245, opened 2020-06-02
Invoke-GraphRequest cmdlet to allow direct requests using underlying HttpClient

finsharp:
Adds an Invoke-GraphRequest cmdlet that enables users to make requests using the underlying authenticated HttpClient.
### Example
`Invoke-GraphRequest -Method GET -Uri https://graph.microsoft.com/beta/me `
`Invoke-GraphRequest -Method GET -Uri /beta/me`

To support non-json responses requires a change in [msgraph-sdk-dotnet-core](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/118) whose compression handler currently doesn't pass along Http Content headers to the next middleware in the pipeline

Address [AB#4897](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4897)


src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+            try
+            {
+                var httpMethod = GetHttpMethod();
+                var requestMessage = new HttpRequestMessage(httpMethod, this.Uri)
+                {
+                    Content = new StringContent(null != Body ? JsonConvert.SerializeObject(Body) : @"{}", Encoding.UTF8,
+                        CoreConstants.MimeTypeNames.Application.Json)
+                };
+                var responseMessage = client.SendAsync(requestMessage).GetAwaiter().GetResult();
+                var responseString = responseMessage.Content.ReadAsStringAsync().GetAwaiter().GetResult();
```
ddyett, line 84: what about non-json responses?

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            base.EndProcessing();
+        }
+
+        protected override void StopProcessing()
+        {
+            base.StopProcessing();
+        }
+
+        private void ValidateParameters()
```
ddyett, line 144: a body is necessary if it's a patch or post.  should there be some validation there to improve messaging in this case?

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            base.ProcessRecord();
+            var authConfig = GraphSession.Instance.AuthContext;
+            var client = HttpHelpers.GetGraphHttpClient(authConfig);
+            try
+            {
+                var httpMethod = GetHttpMethod();
+                var requestMessage = new HttpRequestMessage(httpMethod, this.Uri)
+                {
+                    Content = new StringContent(null != Body ? JsonConvert.SerializeObject(Body) : @"{}", Encoding.UTF8,
```
MIchaelMainer, line 80: It is optional to introduce Newtonsoft. The GraphServiceClient abstracts JsonConvert. You can make the following changes to use the GraphServiceClient JSON converters. Eventually it will use System.Text.Json.

- [ ] **Regarding line 80**

`JsonConvert.SerializeObject(Body)`

becomes

`client.HttpProvider.Serializer.SerializeObject(body)`

- [ ] **Regarding line 85**

`JsonConvert.DeserializeObject<Hashtable>(responseString);`

becomes

`client.HttpProvider.Serializer.DeserializeObject<Hashtable>(responseString);`

- [ ] Remove `using Newtonsoft.Json;`

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
MIchaelMainer: Do we have a work item to track adding non-json requests? Add one if this is not tracked.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
MIchaelMainer: Nevermind. You are using the GraphServiceClient's underlying HttpClient.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            ValidateParameters();
+            base.BeginProcessing();
+        }
+
+        protected override void ProcessRecord()
+        {
+            base.ProcessRecord();
+            var authConfig = GraphSession.Instance.AuthContext;
+            var client = HttpHelpers.GetGraphHttpClient(authConfig);
```
MIchaelMainer, line 74: You already have a GraphHttpClient reference with `this._client`. We are instantiating two instances of HttpClient. Let's use a single one.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+                var responseMessage = client.SendAsync(requestMessage).GetAwaiter().GetResult();
+                var responseString = responseMessage.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                var result = JsonConvert.DeserializeObject<Hashtable>(responseString);
+                if (responseMessage.IsSuccessStatusCode)
+                {
+                    WriteObject(result);
+                }
+                else
+                {
+                    WriteError(new ErrorRecord(new Exception(responseString), Guid.NewGuid().ToString(),
```
MIchaelMainer, line 92: It would also be useful to write out the response headers in case !responseMessage.IsSuccessStatusCode

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        private void ValidateParameters()
+        {
+            if (Uri == null)
+            {
+                ThrowTerminatingError(new ErrorRecord(new ArgumentNullException(nameof(Uri), $"Must specify {Uri}"), Guid.NewGuid().ToString(),
+                    ErrorCategory.InvalidArgument, null));
+            }
+
+            if (Uri.IsAbsoluteUri && _client.BaseAddress.Host != Uri.Host)
```
MIchaelMainer, line 152: Make sure this `_client` is the same one used to make the request. See comment above.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: @MIchaelMainer  Since we are using the underlying HttpClient, i'll make this comment as resolved.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Added

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: image

@MIchaelMainer I am considering this kind of output

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -118,65 +262,704 @@ protected override void ProcessRecord()
...
+        }
+
+        private HttpResponseMessage GetResponse(HttpClient client, HttpRequestMessage request)
+        {
+            if (client == null) throw new ArgumentNullException(nameof(client));
+
+            if (request == null) throw new ArgumentNullException(nameof(request));
+
+            var cancellationToken = _cancelToken.Token;
+            var response = client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
```
MIchaelMainer, line 585: We don't need the cancellation token, it isn't ever cancelled by the user or us. You could add some sort of function or script block parameter that would be cancel the operation on Invoke-GraphRequest. I think we can expect a user will just ctrl+break.  Use the overload without the cancellation token and remove references to the cancellation source.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -118,65 +262,704 @@ protected override void ProcessRecord()
...
+                }
+
+            var bytes = StreamHelper.EncodeToBytes(content, encoding);
+            var byteArrayContent = new ByteArrayContent(bytes);
+            request.Content = byteArrayContent;
+
+            return byteArrayContent.Headers.ContentLength.Value;
+        }
+
+        private void FillRequestStream(HttpRequestMessage request)
```
MIchaelMainer, line 649: general comment: Add a summary to each method so we understand what we are about the read in the function body. The name is decent so I have an idea but this function body has over 100 lines in it so I have to look more than a glance to understand.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -118,65 +262,704 @@ protected override void ProcessRecord()
...
+            {
+                var error = GetValidationError(
+                    "The cmdlet cannot run because the following conflicting parameters are specified: Session and SessionVariable. Specify either Session or SessionVariable, then retry.",
+                    "WebCmdletSessionConflictException");
+                ThrowTerminatingError(error);
+            }
+
             // When PATCH or POST is specified, ensure a body is present
-            if (this.Method == HttpVerb.PATCH || this.Method == HttpVerb.POST && this.Body == null)
+            if (Method == GraphRequestMethod.PATCH || Method == GraphRequestMethod.POST && Body == null)
```
MIchaelMainer, line 850: Body or InFile must not be null.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -118,65 +262,704 @@ protected override void ProcessRecord()
...
+            // Token shouldn't be null when UserProvidedToken is specified
+            if (Authentication == GraphRequestAuthenticationType.UserProvidedToken && string.IsNullOrWhiteSpace(Token))
+            {
+                var error = GetValidationError("AuthenticationCredentialNotSupplied",
+                    "WebCmdletAuthenticationCredentialNotSuppliedException");
+                ThrowTerminatingError(error);
+            }
+
+            // Only Body or InFile can be specified at a time
+            if (Body != null && InFile != null)
```
MIchaelMainer, line 879: optional: You could validate the json if the Body is used by JToken.Parse(string)  and catch jsonreaderexception if it isn't valid. You have that IsJson implemented in the helpers.

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+                return null;
+
+            var sig = contentType.Split(s_contentTypeParamSeparator, 2)[0].ToUpperInvariant();
+            return sig;
+        }
+
+        #endregion Private Helper Methods
+    }
+
+    internal static
```
MIchaelMainer, line 218: multiline -> single line

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+      {
+        throw new ArgumentException("HttpMessageInvalidMediaType", "content");
+      }
+      else
+      {
+        return false;
+      }
+    }
+
+
```
MIchaelMainer, line 174: Comment missing

src/Authentication/Authentication/Helpers/WebResponseHelper.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net.Http;
+using Microsoft.Graph.PowerShell.Authentication.Cmdlets;
+
+namespace Microsoft.Graph.PowerShell.Authentication.Helpers
+{
+    internal class WebResponseHelper
```
MIchaelMainer, line 9: Global. Add documentation to these classes and methods.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -1,87 +1,217 @@
...
+                "Indicates that the cmdlet returns the results, in addition to writing them to a file. Only valid when the OutFile parameter is also used. ")]
+        public SwitchParameter PassThru { get; set; }
+
+        /// <summary>
+        ///     OAuth or Bearer Token to use instead of already acquired token
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = Constants.UserParameterSet,
+            Position = 9,
+            HelpMessage = "OAuth or Bearer Token to use instead of already acquired token")]
+        public string Token { get; set; }
```
MIchaelMainer, line 119: Update the setter to set 

`Authentication = GraphRequestAuthenticationType.UserProvidedToken`

It currently isn't set anywhere and later logic uses this.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: New changes handle non-json responses (images, documents etc)

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            HelpMessage = "Custom User Specified User Agent")]
+        public string UserAgent { get; set; }
+
+        /// <summary>
+        ///     Wait for .NET debugger to attach
+        /// </summary>
+        [Parameter(Mandatory = false,
+            DontShow = true,
+            HelpMessage = "Wait for .NET debugger to attach")]
+        public SwitchParameter Break { get; set; }
```
peombwa, line 197: We shouldn't expose this to customers. This project can already be debugged by pressing F5 and adding breakpoints in Visual Studio. [This](https://docs.poshtools.com/powershell-tools-documentation/chapter1/remote-debugging) is how people generally debug PowerShell scripts.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
peombwa: We should cancel the `CancellationTokenSource` in the `StopProcessing` override method. `StopProcessing` is the method that's called on `ctrl+break`. See the implementation here https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Authentication/Authentication/Cmdlets/ConnectGraph.cs#L140

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            HelpMessage = "Infer Download FileName")]
+        public SwitchParameter InferOutFileName { get; set; }
+
+        /// <summary>
+        ///     Gets or sets the InFile property.
+        /// </summary>
+        [Parameter]
+        [Parameter(Mandatory = false, ParameterSetName = Constants.UserParameterSet,
+            Position = 7,
+            HelpMessage = "Infile to Send in the Request")]
```
zengin, line 100: nit: "Input file to send in the request"

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            Position = 5,
+            HelpMessage = "Output file where the response body will be saved")]
+        public string OutFile { get; set; }
+
+        /// <summary>
+        ///     Infer Download FileName from ContentDisposition Header
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = Constants.UserParameterSet,
+            Position = 6,
+            HelpMessage = "Infer Download FileName")]
```
zengin, line 91: nit: Infer output file name

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            Mandatory = false,
+            HelpMessage = "Response Status Code Variable")]
+        public string StatusCodeVariable { get; set; }
+
+        /// <summary>
+        ///     Gets or sets whether to skip checking HTTP status for error codes.
+        /// </summary>
+        [Parameter(Position = 16, ParameterSetName = Constants.UserParameterSet, Mandatory = false, HelpMessage = "Skip Checking Http Errors")]
+        public virtual SwitchParameter SkipHttpErrorCheck { get; set; }
+        /// <summary>
```
zengin, line 175: nit: empty line

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            return errorRecord;
+        }
+
+        private void ReportRequestStatus(HttpRequestMessage requestMessage)
+        {
+            long requestContentLength = 0;
+            if (requestMessage.Content != null)
+            {
+                requestContentLength = requestMessage.Content.Headers.ContentLength.Value;
+            }
```
zengin, line 304: nit: `long requestContentLength = requestMessage.Content?.Headers.ContentLength.Value ?? 0;`

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+
+        private static string FormatDictionary(IDictionary content)
+        {
+            if (content == null)
+                throw new ArgumentNullException(nameof(content));
+
+            var bodyBuilder = new StringBuilder();
+            foreach (string key in content.Keys)
+            {
+                if (0 < bodyBuilder.Length) bodyBuilder.Append("&");
```
zengin, line 334: STYLE: I see three different versions of one-statement blocks in this code change:
```
if (condition)
  statement
```

```
if (condition)
{
  statement
}
```

```
if (condition) statement
```
Please use one of them for consistency in style.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+
+            return bodyBuilder.ToString();
+        }
+
+        private HttpRequestMessage GetRequest(Uri uri)
+        {
+            var requestUri = PrepareUri(uri);
+            var httpMethod = GetHttpMethod(Method);
+            // create the base WebRequest object
+            var request = new HttpRequestMessage(httpMethod, requestUri);
```
zengin, line 354: `using var` ?

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            if (GraphRequestSession.Headers.Count > 0)
+            {
+                GraphRequestSession.ContentHeaders.Clear();
+                foreach (var entry in GraphRequestSession.Headers)
+                {
+                    if (HttpKnownHeaderNames.ContentHeaders.Contains(entry.Key))
+                    {
+                        GraphRequestSession.ContentHeaders.Add(entry.Key, entry.Value);
+                    }
+                    else
```
zengin, line 366: `else if`

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+                            request.Headers.Add(entry.Key, entry.Value);
+                        }
+                    }
+                }
+            }
+
+            // Set 'Transfer-Encoding: chunked' if 'Transfer-Encoding' is specified
+            if (GraphRequestSession.Headers.ContainsKey(HttpKnownHeaderNames.TransferEncoding))
+            {
+                request.Headers.TransferEncodingChunked = true;
```
zengin, line 383: What happens if the header is set to `false` explicitly?

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            }
+
+            // Set 'User-Agent' if WebSession.Headers doesn't already contain it
+            if (GraphRequestSession.Headers.TryGetValue(HttpKnownHeaderNames.UserAgent, out var userAgent))
+            {
+                GraphRequestSession.UserAgent = userAgent;
+            }
+            else
+            {
+                if (SkipHeaderValidation)
```
zengin, line 393: `else if`

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+
+            return request;
+        }
+
+        private Uri PrepareUri(Uri uri)
+        {
+            // before creating the web request,
+            // preprocess Body if content is a dictionary and method is GET (set as query)
+            LanguagePrimitives.TryConvertTo(Body, out IDictionary bodyAsDictionary);
+            if (bodyAsDictionary != null && Method == GraphRequestMethod.GET)
```
zengin, line 412: PERF: if you change these two lines into:
`if (Method == GraphRequestMethod.GET && LanguagePrimitives.TryConvertTo(Body, out IDictionary bodyAsDictionary))`

you will avoid converting the body into dictionary if the method is not `GET`.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+                        var fileName = response.Content.Headers.ContentDisposition.FileNameStar;
+                        var fullFileName = QualifyFilePath(fileName);
+                        WriteVerbose(string.Format(CultureInfo.InvariantCulture, "Inferred File Name {0} Saving to {1}", fileName, fullFileName));
+                        StreamHelper.SaveStreamToFile(baseResponseStream, fullFileName, this,
+                            _cancelToken.Token);
+                    }
+                }
+                else
+                {
+                    WriteVerbose("Could not Infer File Name");
```
zengin, line 518: nit: Could not infer file name

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            if (GraphRequestSession != null && SessionVariable != null)
+            {
+                var error = GetValidationError(
+                    "The cmdlet cannot run because the following conflicting parameters are specified: Session and SessionVariable. Specify either Session or SessionVariable, then retry.",
+                    "WebCmdletSessionConflictException");
+                ThrowTerminatingError(error);
+            }
+
+            // When PATCH or POST is specified, ensure a body is present
+            if (Method == GraphRequestMethod.PATCH || Method == GraphRequestMethod.POST && Body == null)
```
zengin, line 850: This logic doesn't do what is mentioned in the comment above. `&&` comes higher in the order of precedence against `||`. If Method is `PATCH`, the other conditions will not be checked.

```suggestion
            if ((Method == GraphRequestMethod.PATCH || Method == GraphRequestMethod.POST) && Body == null)
```

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+        private ErrorRecord GetValidationError(string msg, string errorId)
+        {
+            var ex = new ValidationMetadataException(msg);
+            var error = new ErrorRecord(ex, errorId, ErrorCategory.InvalidArgument, this);
+            return error;
+        }
+
+        private ErrorRecord GetValidationError(string msg, string errorId, params object[] args)
+        {
+            msg = string.Format(CultureInfo.InvariantCulture, msg, args);
```
zengin, line 949: Q: what is the reasoning behind using InvariantCulture here and CurrentCulture in other places?

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+                var statusCode = WebResponseHelper.GetStatusCode(response);
+                var statusDescription = WebResponseHelper.GetStatusDescription(response);
+                raw.AppendFormat("{0} {1} {2}", protocol, statusCode, statusDescription);
+                raw.AppendLine();
+            }
+
+            HttpHeaders[] headerCollections =
+            {
+                response.Headers,
+                response.Content == null ? null : response.Content.Headers
```
zengin, line 85:
```suggestion
                response.Content?.Headers
```

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+            };
+
+            foreach (var headerCollection in headerCollections)
+            {
+                if (headerCollection == null)
+                {
+                    continue;
+                }
+
```
zengin, line 94:
```suggestion
            foreach (var headerCollection in headerCollections.Where(collection => collection != null))
            {
```

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+            // Further content type analysis is available on Windows
+            if (Platform.IsWindows && !isText)
+            {
+                // Media types registered with Windows as having a perceived type of text, are text
+                using (var contentTypeKey =
+                    Registry.ClassesRoot.OpenSubKey(@"MIME\Database\Content Type\" + contentType))
+                {
+                    if (contentTypeKey != null)
+                    {
```
zengin, line 171: inline this check in the next line as `contentTypeKey?.GetValue...`

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+            if (string.IsNullOrEmpty(contentType))
+                return false;
+
+            // RFC 3023: Media types with the suffix "+xml" are XML
+            var isXml = contentType.Equals("application/xml", StringComparison.OrdinalIgnoreCase)
+                        || contentType.Equals("application/xml-external-parsed-entity",
+                            StringComparison.OrdinalIgnoreCase)
+                        || contentType.Equals("application/xml-dtd", StringComparison.OrdinalIgnoreCase);
+
+            isXml |= contentType.EndsWith("+xml", StringComparison.OrdinalIgnoreCase);
```
zengin, line 202:
```suggestion
            var isXml = contentType.StartsWith("application/xml", StringComparison.OrdinalIgnoreCase)
                || contentType.EndsWith("+xml", StringComparison.OrdinalIgnoreCase);
```

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            var respVerboseMsg = StringFormatCurrentCulture("received {0}-byte response of content type {1}",
+                responseMessage.Content.Headers.ContentLength,
+                contentType);
+            WriteVerbose(respVerboseMsg);
+        }
+
+        private static string StringFormatCurrentCulture(string format, params object[] args)
+        {
+            return string.Format(CultureInfo.CurrentCulture, format, args);
+        }
```
zengin, line 324: can be replaced with the function you are defining in `StringUtil`.

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+            // the correct type for JSON content, as specified in RFC 4627
+            var isJson = contentType.Equals("application/json", StringComparison.OrdinalIgnoreCase);
+
+            // add in these other "javascript" related types that
+            // sometimes get sent down as the mime type for JSON content
+            isJson |= contentType.Equals("text/json", StringComparison.OrdinalIgnoreCase)
+                      || contentType.Equals("application/x-javascript", StringComparison.OrdinalIgnoreCase)
+                      || contentType.Equals("text/x-javascript", StringComparison.OrdinalIgnoreCase)
+                      || contentType.Equals("application/javascript", StringComparison.OrdinalIgnoreCase)
+                      || contentType.Equals("text/javascript", StringComparison.OrdinalIgnoreCase);
```
zengin, line 148: to avoid repeating `StringComparison.OrdinalIgnoreCase`, you can create a `switch` block on `contentType.ToLower(CultureInfo.InvariantCulture)`

src/Authentication/Authentication/Helpers/ContentHelper.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        internal static
+            string
+            Format(string formatSpec, params object[] o)
+        {
+            return string.Format(CultureInfo.CurrentCulture, formatSpec, o);
+        }
+
+        internal static
```
zengin, line 251: Where are the following three functions used?

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Graph.PowerShell.Authentication.Helpers
+{
+  /// <summary>
```
zengin, line 16: STYLE: Please switch back to 4 spaces for indentation for consistency.

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+    private static readonly HashSet<string> _neverSerializedHeaderFields = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+    {
+      "SdkVersion",
+      "FeatureFlag",
+      "Authorization",
+      "Cache-Control",
+      "Transfer-Encoding",
+      "Duration",
+      "Strict-Transport-Security",
+      "Date"
```
zengin, line 59: use constants from known header names

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+      Headers.ContentType = new MediaTypeHeaderValue(DefaultMediaType);
+      Headers.ContentType.Parameters.Add(new NameValueHeaderValue(MsgTypeParameter, DefaultResponseMsgType));
+
+      InitializeStreamTask();
+    }
+
+    private HttpContent Content
+    {
+      get { return HttpRequestMessage != null ? HttpRequestMessage.Content : HttpResponseMessage.Content; }
+    }
```
zengin, line 96: Q: Is this a convention of some sort?

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+    /// </summary>
+    /// <param name="content">The content to validate.</param>
+    /// <param name="isRequest">if set to <c>true</c> if the content is either an HTTP Request or an HTTP Response.</param>
+    /// <param name="throwOnError">Indicates whether validation failure should result in an <see cref="Exception"/> or not.</param>
+    /// <returns><c>true</c> if content is either an HTTP Request or an HTTP Response</returns>
+    internal static bool ValidateHttpMessageContent(HttpContent content, bool isRequest, bool throwOnError)
+    {
+      if (content == null)
+      {
+        throw new ArgumentNullException("content");
```
zengin, line 124:
```suggestion
        throw new ArgumentNullException(nameof(content));
```

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+          {
+            throw new ArgumentException("HttpMessageInvalidMediaType", "content");
+          }
+          else
+          {
+            return false;
+          }
+        }
+
+        foreach (NameValueHeaderValue parameter in contentType.Parameters)
```
zengin, line 142: with the current flow, there is a way to reach here with `contentType == null`

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+            }
+
+            return true;
+          }
+        }
+      }
+
+      if (throwOnError)
+      {
+        throw new ArgumentException("HttpMessageInvalidMediaType", "content");
```
zengin, line 166: nameof

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+
+
+    public static string UnquoteToken(string token)
+    {
+      if (String.IsNullOrWhiteSpace(token))
+      {
+        return token;
+      }
+
+      if (token.StartsWith("\"", StringComparison.Ordinal) && token.EndsWith("\"", StringComparison.Ordinal) && token.Length > 1)
```
zengin, line 182: PERF: make `token.Length > 1` first condition

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
```diff
@@ -0,0 +1,359 @@
...
+        throw new ArgumentException("HttpMessageInvalidMediaType", "content");
+      }
+      else
+      {
+        return false;
+      }
+    }
+
+
+    public static string UnquoteToken(string token)
```
zengin, line 175: If the input string guaranteed **not** to have repeated quotes at the beginning and at the end, the whole function can be replaced with `token.Trim('"')`

src/Authentication/Authentication/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,178 @@
...
+                do
+                {
+                    // check for a charset attribute on the meta element to override the default.
+
+                    var localEncoding = Encoding.UTF8;
+                    responseStream.Seek(0, SeekOrigin.Begin);
+                    content = StreamToString(responseStream, localEncoding);
+                    // report the encoding used.
+                    encoding = localEncoding;
+                } while (false);
```
zengin, line 74: Q: What is the purpose of having a loop here? This loop seems to be guaranteed to execute once.

src/Authentication/Authentication/Helpers/StreamHelper.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            return content;
+        }
+
+        private static string StreamToString(Stream stream, Encoding encoding)
```
zengin, line 80: There are a few ways to achieve this using the standard library, one way:

```
using StreamReader reader = new StreamReader(stream, encoding);
return reader.ReadToEnd();
```

src/Authentication/Authentication/Helpers/WebResponseHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        internal static string GetCharacterSet(HttpResponseMessage response)
+        {
+            var characterSet = response.Content.Headers.ContentType.CharSet;
+            return characterSet;
+        }
```
zengin, line 15: consider making this an extension method by changing the argument to `(this HttpResponseMessage response)`

src/Authentication/Authentication/Helpers/WebResponseHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+            var statusDescription = response.StatusCode.ToString();
+            return statusDescription;
+        }
+
+        internal static bool IsText(HttpResponseMessage response)
+        {
+            // ContentType may not exist in response header.
+            var contentType = response.Content.Headers.ContentType?.MediaType;
+            return ContentHelper.IsText(contentType);
+        }
```
zengin, line 64: can be extension methods

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Tooling issue. Resharper seems confused. Will correct.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: In the constructor its set to Default.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+
+        /// <summary>
+        ///     Http Method
+        /// </summary>
+        [Parameter(ParameterSetName = Constants.UserParameterSet,
+            Position = 1,
+            Mandatory = true,
+            HelpMessage = "Http Method")]
+        public GraphRequestMethod Method { get; set; }
```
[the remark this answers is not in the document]
finsharp: Its already set in the constructor to GET. Mandatory requirement is what I'll remove.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: C# 8.0 feature simplifying using statements. 
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

peombwa: We should not be using C# 8 in this project. Some C# 8 features depend on .NET Standard 2.1 and .NET Core 3.x, see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version.

To maintain consistency with the generated PS module projects, we should always use C# 7.1.

zengin: I saw `using var` being used on other files, so I assumed that we can use C# 8. @finsharp, per @peombwa's comment, please revert them in `InvokeGraphRequest.cs` as well.

`IDisposable` objects should dispose properly. Even if you can't have `using var`, you can always have `using` blocks.

finsharp: Will revert to 7.1

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Corrected. Using current culture now consistently.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Moved to StringUtil

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
finsharp: The formatter can format both requests and responses, so here is to return whichever is available.

zengin: I can't think of a scenario where the consumer of this code would ask for `.Content` property without knowing whether they are going to get the contents of request or response. That sounds like an arbitrary behavior. But maybe I am missing an important detail.

Do you mind clarifying with a call flow for me to better understand the coupling of "`Content` can return response content if an only if request is null"?

If this behavior is not intended, I would suggest having two different properties, e.g. `RequestContent` and `ResponseContent`.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+        /// <summary>
+        ///     Uri to call using the Graph HttpClient can be segments such as /beta/me
+        ///     or fully qualified url such as https://graph.microsoft.com/beta/me
+        /// </summary>
+        [Parameter(ParameterSetName = Constants.UserParameterSet,
+            Position = 2,
+            Mandatory = true,
+            HelpMessage =
+                "Uri to call can be segments such as /beta/me or fully qualified https://graph.microsoft.com/beta/me")]
+        public Uri Uri { get; set; }
```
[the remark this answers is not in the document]
finsharp: image

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
finsharp: HttpMessageFormatter, can format both requests and responses. It has two constructors, one takes a request and the other takes a response. So when a Request is passed in (this field is private though), HttpContent will point to the the Request content.

src/Authentication/Authentication/Helpers/HttpMessageFormatter.cs
finsharp: If contentType is null, and throwOnError is true, an exception is thrown.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -1,87 +1,217 @@
...
+            HelpMessage = "Output file where the response body will be saved")]
+        public string OutFile { get; set; }
+
+        /// <summary>
+        ///     Infer Download FileName from ContentDisposition Header
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = Constants.UserParameterSet,
+            Position = 6,
+            HelpMessage = "Infer Download FileName")]
+        public SwitchParameter InferOutFileName { get; set; }
```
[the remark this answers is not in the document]
finsharp: Yes, in verbose mode, it informs the user of the issue.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Since body can be json, byte array, a dictionary or a stream, just checking for null is enough.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -0,0 +1,965 @@
...
+            HelpMessage = "Custom Graph Request Session")]
+        public GraphRequestSession GraphRequestSession { get; set; }
+
+        /// <summary>
+        ///     Custom User Specified User Agent
+        /// </summary>
+        [Parameter(Mandatory = false,
+            Position = 18, ParameterSetName = Constants.UserParameterSet,
+            HelpMessage = "Custom User Specified User Agent")]
+        public string UserAgent { get; set; }
```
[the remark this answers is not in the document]
finsharp: Both ways work. The field is not mandatory.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: It cannot be set to false, since that's now an allowed value.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding
Allowed Values are:- 

1. - Transfer-Encoding: chunked
2. - Transfer-Encoding: compress
3. - Transfer-Encoding: deflate
4. - Transfer-Encoding: gzip
5. - Transfer-Encoding: identity

The property request.Headers.TransferEncodingChunked, just causes (1) to be emitted.

src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
finsharp: Please elaborate...Looks okay to me. Is there something I am missing?

src/Authentication/Authentication/Helpers/StringUtil.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return string.Format(CultureInfo.CurrentCulture, formatSpec, args);
+        }
+
+        /// <summary>
+        ///     Formats a Dictionary into a UrlEncoded string
+        /// </summary>
+        /// <param name="content"></param>
+        /// <returns></returns>
+        internal static string FormatDictionary(this IDictionary content)
```
zengin, line 35: NIT: If it makes sense to take a dependency on `Microsoft.AspNetCore.WebUtilities`, you can then use `QueryHelpers.AddQueryString`.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.queryhelpers.addquerystring?view=aspnetcore-3.1

finsharp: @peombwa What are your thoughts on introducing that dependency?

peombwa: @finsharp the library is not shipped with PowerShell, and we may have to ship the dll with our SDK. That means adding an extra 55KB to the SDK. See https://github.com/Azure/azure-powershell/issues/7921. 

We can omit it for now since we have this method as an alternative. Do we see us needing to use other methods in [microsoft.aspnetcore.webutilities](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities?view=aspnetcore-2.0)?

finsharp: For now I don't think we will need the extra functionality. Will omit it for now till the need arises

zengin: Thanks for the context @peombwa. I agree on not introducing a dependency in this case.

PR #254, opened 2020-06-09
Multi Profile Support - Part 1

peombwa:
This PR is divided into 2 parts, part 1 of this PR closes #72, [AB#4769](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4769) by proposing the following changes:
1. Adds `tools/GenerateProfiles.ps1` script which will generate profiles definition files from downloaded OpenAPI files.
   - Profile definitions are a mapping of operation paths and API versions. e.g. [Users.User v1.0 profile definition](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/po/mutliProfileProjStructure/profiles/Users.User/definitions/v1.0-beta.md).
   - Each module can have more than one profile.
   - Each module will have a single profile `readme.md` that imports all supported profiles, and input-files to the profile's OpenAPI doc e.g. [Users.User module profile readme](https://github.com/microsoftgraph/msgraph-sdk-powershell/tree/po/mutliProfileProjStructure/profiles/Users.User).
2. Updates `tools/GenerateModules.ps1` to add supported profiles to a modules manifest (.psd1). This will be used by `Get-MgProfile -ListAvailable` to get all supported profiles. e.g. [Users.User module manifest](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/po/mutliProfileProjStructure/src/Analytics/Analytics/Microsoft.Graph.Analytics.psd1#L110).
3. Adds `src/Authentication/Authentication/Cmdlets/GetMgProfile.cs` and `src/Authentication/Authentication/Cmdlets/SelectMgProfile.cs` to manage supported Microsoft Graph Profiles.

The profile definitions will be generated, and consumed to generate modules as show below:
image

##### Profiles in actions
Profiles in action

[Part 2](#255 ) of this PR has the profiles generated by `tools/GenerateProfiles.ps1`, an updated project structure, and a complete regeneration of all modules using v1.0 & beta OpenAPI docs.

Issue #198, opened 2020-05-04
Users.Actions cmdlets returning unrecognized response


Windos:
To be clear I'm not sure how many Users.Actions cmdlets do this, but when I run `Move-MgUserMessage` the response is "Created" (it looks like it is creating a job to be processed some moments/minutes later) which results in an error.

![Unrecognized response exception](https://user-images.githubusercontent.com/6955786/81009154-3438a980-8ea8-11ea-9a07-772010af9630.png)

Code to repro this is the same as what I included in #187

Windos:
Just touching base to say this was retested in 0.5.1 and can confirm issue is still reproducable.

MIchaelMainer:
Investigate custom directive to change how we process responses with unspecified (per open API 
 spec) response status. related to  #182

peombwa:
Closing since this has been fixed. Use ` Install-Module Microsoft.Graph -AllowClobber -Force` to install the latest version of the SDK.

PR #272, opened 2020-06-23
Proper Success Response Handling

### LINKED ISSUE #110: Update-MgUserEvent throws expeption but is running fine
When running  `Update-MgUserEvent -UserId X -EventId X -Subject  X`  i get the folling error but subject gets changed as expected.

```
Update-MgUserEvent : The server responded with an unrecognized response, Status: OK
At line:11 char:10
+          Update-MgUserEvent -UserId $Mail -EventId $_.Id -Subject $_. ...
+          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: ({ UserId = surf...oftGraphEvent }:<>f__AnonymousType513) [Update-MgUserEvent_UpdateExpanded], RestException1
    + FullyQualifiedErrorId : OK,Microsoft.Graph.PowerShell.Cmdlets.UpdateMgUserEvent_UpdateExpanded
```

### LINKED ISSUE #271: Allow customers to iterate collection of pages based on their PageSize/top.
When `-PageSize` and `-All` are used together, we should fetch all items in the page size specified by the customer. The current paging design defaults to fetching the total number of items in `-PageSize`, and does not iterate through the collection of pages.

This is a follow-up to #266

peombwa:
This PR closes #110 and #198 by adding  `OverrideOnDefault` method on `PSCmdlet`. It will ensure cmdlets won't fail when a service returns a success code that's not defined in the OpenAPI coc. The  `OverrideOnDefault` will be added to the generated cmdlet at generation time by the directive in [readme.graph.md](https://github.com/microsoftgraph/msgraph-sdk-powershell/compare/po/successResponse?expand=1#diff-f6090dc5a1d5d42d140553e4691c12adR391).

The generated code will call `OverrideOnDefault` after considering the status code defined in the OpenAPI doc before giving a response a blanket success if its status code is 2xx as shown below.
image
image

The PR also closes #271 by allowing customers to specify a custom `-PageSize` when iterating through collection of pages as described in https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/275.
image

#### Checklist Before Merge
- [ ] Merge #254 and #255 .
- [ ] Retest the directive with the new project structure with profiles.

[AB#5066](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5066)
[AB#5067](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5067)


KirkMunro:
One thing to consider with this as a fix for #271:

Making `-PageSize` properly control the page size while allowing users to retrieve all records is the right thing to do.

But `-PageSize` currently has `-Limit` and `-Top` as aliases, and neither of those aliases are appropriate for `-PageSize` because they are about limiting the overall number of records returned, not about controlling the size of the pages used to retrieve those records.

I think `-Limit` should be broken out into its own parameter with `-Top` as an alias to work like `-PageSize` did before this change.

Related: #275

tools/Custom/ListCmdlet.cs
```diff
@@ -84,11 +101,8 @@ public bool ShouldIteratePages(global::System.Collections.Generic.Dictionary<str
         {
             iteratedPages++;
             totalFetchedItems += itemsCount;
-            if ((boundParameters.ContainsKey("All") && !boundParameters.ContainsKey("PageSize")) ||
-                (boundParameters.ContainsKey("PageSize") && totalFetchedItems < originalPageSize))
-                return true;
-            else
-                return false;
+
+            return limit == default || totalFetchedItems < limit;
```
ddyett, line 105: may be helpful to add some quick comments.  For example 'we iterate if there's no top/limit parameter or if there are more items to fetch'

tools/Custom/ListCmdlet.cs
```diff
@@ -56,26 +52,47 @@ public partial class ListCmdlet : global::System.Management.Automation.PSCmdlet
...
+            }
+
+            // Move `-Top` parameter to `limit` parameter.
+            if (invocationInfo.BoundParameters.ContainsKey("Top"))
+            {
+                limit = top;
+            }
+
+            // Explicitly set `-PageSize` as our $top parameter.
+            invocationInfo.BoundParameters["Top"] = invocationInfo.BoundParameters.ContainsKey("PageSize") ? PageSize : DefaultPageSize;
```
ddyett, line 84: why do we need to do this?

tools/Custom/ListCmdlet.cs
```diff
@@ -56,26 +52,47 @@ public partial class ListCmdlet : global::System.Management.Automation.PSCmdlet
...
+            top = (int)invocationInfo.BoundParameters["Top"];
+
+            if (limit != default)
+            {
+                requiredPages = limit / top;
```
ddyett, line 89: this line just looks weird as you'd expect limit/pageSize

peombwa: That's the intent, let me clean this up.

tools/Custom/ListCmdlet.cs
peombwa: We need it in order to construct a URL with a `$top` query parameter in the generated cmdlets. `PageSize` is a just a runtime parameter and is never serialized. Top/Limit on the other hand is a query parameter that's defined in the OpenAPI doc, and it will be serialized as `$top` query parameter.
``` cs
await this.Client.UsersUserListUser(this.InvocationInformation.BoundParameters.ContainsKey("Top") ? Top : default(int?), this.InvocationInformation.BoundParameters.ContainsKey("Skip") ? Skip : default(int?)...
 ```

peombwa:
Due to COGs concerns, we will not fetch all records from the service by default unless the user explicitly opts-in via `-All` parameter. See discussion [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/275#issuecomment-653796289).

peombwa:
@ddyett, @darrelmiller, I've updated the PR to address the COGs concerns with fetching all pages by default. With the update made in https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/272/commits/d4f87979970891891870e4dcd3bb6c8fe155a9d5, we will fetch a single page of 100 items by default. `-All` parameter can then be used to fetch all items, and `-PageSize` to specify your desired page size. The new behavior will be as follows:
 
image
cc @KirkMunro

PR #288, opened 2020-07-10
0.7.0 fix

### LINKED ISSUE #279: Important: Move token caching off disk and just keep it cached in memory in the current process.
The way this module bundle is written today, you write an encrypted token to disk (on Windows) or to a keychain or similar on Mac/Linux). This is way too limiting.

With the current design, if I have multiple scripts running at the same time on the same system that each independently create a connection to Graph using the same app registration, do their work, and then disconnect from Graph, each in their own process, it is possible for one of those scripts to fail when the cache file is in use by another process. We just hit this with our application that uses this module bundle, and it's a serious issue. We run scheduled and ad-hoc workloads on a system that has the Graph module installed, and one workload cannot and should not be concerned if another unrelated workload is running at the same time against the same tenant with the same app registration. That's simply a showstopper for using automation with Graph and this module bundle.

Further, if I'm managing my script workloads in containers or serverless environments where I don't necessarily have access to disk and where I don't need access to disk, this won't fly either.

Please take on a task in your roadmap to move this caching off disk, and just keep the encrypted token cached in memory in the current process instead. Failing that, I suppose if you must cache on disk, at least do it in a way that multiple independent processes using the same app id can connect to/disconnect from Graph without tripping over each other.

This issue is so limiting, I think you really should consider releasing a 0.7.1 release with a hotfix for it. Otherwise, the Graph module bundle just won't scale to meet real-world automation needs, and 0.9.0 hasn't even been assigned a publicly visible due date yet.

peombwa:
This PR closes #279 by making the following changes:
- Default to process cache for app only auth i.e. scripted experience, and user cache file for delegated authentication, device code.
- Adds`-ContextScope` enum parameter to `Connect-Graph` cmdlet. The value can be `Process` or `CurrentUser`. This allows customers to opt-out of the default experience.
- Updates cache locking mechanism to one that's optimized for reads. We expect n-reads and 1 write per hour to the token cache.
- Adds unit tests to cover process cache.

###### Default 
| Call | App Type  | Context Scope  |
|---|---|---|
|  `Connect-Graph -Scopes "User.Read"` | Delegated  | `CurrentUser`  |
| `Connect-Graph -ClientId {Client_ID} -TenantId {Tenant_Id} -CertificateThumbprint {Cert_Thumbprint}`  | App Only  | `Process`   |

###### With `-ContextScope`
| Call | App Type  | Context Scope  |
|---|---|---|
|  `Connect-Graph -ContextScope Process -Scopes "User.Read"` | Delegated  | `Process `  |
| `Connect-Graph -ContextScope CurrentUser -ClientId {Client_ID} -TenantId {Tenant_Id} -CertificateThumbprint {Cert_Thumbprint}`  | App Only  | `CurrentUser `   |


src/Authentication/Authentication/Common/GraphSession.cs
```diff
@@ -21,6 +21,19 @@ public class GraphSession : IGraphSession
...
+
+        /// <summary>
+        /// Gets or Sets a session based token.
+        /// This returns an empty byte[] when token is not present.
+        /// </summary>
+        public byte[] Token
+        {
+            get { return token ?? new byte[0]; }
+            set { token = value; }
+        }
```
zengin, line 34: Is this semantically equivalent to the following?
`public byte[] Token { get; set; } = new byte[0];`

src/Authentication/Authentication/TokenCache/TokenCacheStorage.cs
```diff
@@ -45,59 +53,73 @@ public static byte[] GetToken(string appId)
...
+                    WindowsTokenCache.SetToken(authContext.ClientId, accessToken);
+                }
+                else if (Helpers.OperatingSystem.IsMacOS())
+                {
+                    MacTokenCache.SetToken(authContext.ClientId, accessToken);
+                }
+                else
+                {
+                    LinuxTokenCache.SetToken(authContext.ClientId, accessToken);
+                }
```
zengin, line 87: since this pattern seems to be repeating in at least 3 different places, do we have an interface where all three token caches implement, so that we can do something like:
```
var cache = Helpers.GetOSCache();
cache.SetToken(clientId, accessToken)
```

also for `DeleteToken` and `GetToken` respectively?

src/Authentication/Authentication/Common/GraphSession.cs
peombwa: It's different, we always want the value of `Token` to be an empty byte[] when no value is present, or when value is null. `public byte[] Token { get; set; } = new byte[0];` only considers the default case, but it allows for null.

src/Authentication/Authentication/TokenCache/TokenCacheStorage.cs
peombwa: This is a great suggestion, I'll add an interface for this.

peombwa: On problem though, C# 7 doesn't allow static interfaces.

zengin: Singleton non-static classes can achieve the same thing I presume, but feel free to not take that change if that means too much refactoring for the scope of this fix.

src/Authentication/Authentication/Cmdlets/ConnectGraph.cs
```diff
@@ -63,28 +65,33 @@ protected override void ProcessRecord()
                 CancellationTokenSource cts = new CancellationTokenSource(authTimeout);
                 cancellationToken = cts.Token;
 
-                authConfig.AuthType = AuthenticationType.Delegated;
-                authConfig.Scopes = Scopes ?? new string[] { "User.Read" };
+                authContext.AuthType = AuthenticationType.Delegated;
+                authContext.Scopes = Scopes ?? new string[] { "User.Read" };
+                authContext.ContextScope = ContextScope.CurrentUser;
```
ddyett, line 70: don't think you need this line.

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -3,75 +3,95 @@
...
     internal static class AuthenticationHelpers
     {
-        private static readonly object FileLock = new object();
+        static ReaderWriterLockSlim _cacheLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
 
-        internal static IAuthenticationProvider GetAuthProvider(IAuthContext authConfig)
+        internal static IAuthenticationProvider GetAuthProvider(IAuthContext authContext)
         {
-            if (authConfig.AuthType == AuthenticationType.Delegated)
+            if (authContext is null)
```
ddyett, line 20: use brackets to avoid future mistakes.

src/Authentication/Authentication/Extensions/PSCmdletExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph.PowerShell.Authentication.Extensions
+{
+    using System.Management.Automation;
+    internal static class PSCmdletExtensions
+    {
+        /// <summary>
+        /// Determines is a parameter has been provided by the user.
```
ddyett, line 11:
```suggestion
        /// Determines if a parameter has been provided by the user.
```

src/Authentication/Authentication/TokenCache/TokenCacheStorage.cs
```diff
@@ -45,59 +53,73 @@ public static byte[] GetToken(string appId)
...
             }
 
-            if (Helpers.OperatingSystem.IsWindows())
+            if (authContext.ContextScope == ContextScope.Process)
             {
-                WindowsTokenCache.SetToken(appId, accessToken);
+                GraphSession.Instance.Token = accessToken;
             }
-            else if (Helpers.OperatingSystem.IsMacOS())
+            else if(authContext.ContextScope == ContextScope.CurrentUser)
```
ddyett, line 74:
```suggestion
            else if (authContext.ContextScope == ContextScope.CurrentUser)
```

src/Authentication/Authentication/Cmdlets/ConnectGraph.cs
peombwa: Good catch!

PR #297, opened 2020-07-22
IGR Error Fixes

### LINKED ISSUE #296: Invoke-GraphRequest fails for the following scenarios.
1. When a GET call with a query parameter is made to `Invoke-GraphRequest`, it throws a null exception at [InvokeGraphRequest.cs#L352](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs#L352) because we are calling `.FormatDictionary()` on a null `bodyAsDictionary`. We should check if `Body` is null before attempting to convert it to an `IDictionary`.
**Repro** - `Invoke-GraphRequest -Uri 'https://graph.microsoft.com/beta/groups?$count=true'`.

2. When a PowerShell hashtable with values is passed to `-Headers` parameter, a type conversion exception is thrown. This is because the parameter is written to accept `IDictionary<string,string>`, and not `IDictionary`. A non-generic dictionary maps to a PowerShell hashtable.
**Repro** - `Invoke-GraphRequest -Uri 'https://graph.microsoft.com/beta/groups?$count=true' -Headers @{"ConsistencyLevel" = "eventual"}`

finsharp:
Ensure that Connect-Graph has been prior called before IGR is called.
Change Headers to non-generic Dictionary.
Traverse nested json.
Add resource messages for error messages.
Provide ability to pass relative URL as well as query params.

fixes #296

### DIFF
--- a/src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
+++ b/src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
@@ -293,11 +293,12 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
         /// <summary>
         ///     Compose a request, setting Uri and Headers.
         /// </summary>
+        /// <param name="httpClient"></param>
         /// <param name="uri"></param>
         /// <returns></returns>
-        private HttpRequestMessage GetRequest(Uri uri)
+        private HttpRequestMessage GetRequest(HttpClient httpClient, Uri uri)
         {
-            var requestUri = PrepareUri(uri);
+            var requestUri = PrepareUri(httpClient, uri);
             var httpMethod = GetHttpMethod(Method);
             // create the base WebRequest object
             var request = new HttpRequestMessage(httpMethod, requestUri);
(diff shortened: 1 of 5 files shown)


src/Authentication/Authentication/Cmdlets/InvokeGraphRequest.cs
```diff
@@ -25,13 +25,15 @@ public class InvokeGraphRequest : PSCmdlet
...
         private readonly InvokeGraphRequestUserAgent _graphRequestUserAgent;
+        private readonly GraphSession _graphSession;
         private string _originalFilePath;
 
         public InvokeGraphRequest()
         {
             _cancellationTokenSource = new CancellationTokenSource();
             _graphRequestUserAgent = new InvokeGraphRequestUserAgent(this);
             Authentication = GraphRequestAuthenticationType.Default;
+            _graphSession = GraphSession.Instance;
```
peombwa, line 36: `GraphSession.Instance` is a static readonly property. There is no need to store it a variable, it also has the benefit of being thread safe.

finsharp: What conditions make AuthContext null?

src/Authentication/Authentication/Properties/Resources.resx
```diff
@@ -196,4 +196,13 @@
   <data name="InferFileNameOutFilePathConflict" xml:space="preserve">
     <value>The cmdlet cannot run because the following conflicting parameters are specified: {0} and {1}. Specify either {0} or {1} then retry.</value>
   </data>
+  <data name="NotConnectedToGraphException" xml:space="preserve">
+    <value>The cmdlet cannot run because Authentication is set to {0} and Connect-Graph was not called. Invoke 'Connect-Graph' or specify Authentication to be 'UserProvidedToken' and Provide a Token then retry</value>
```
peombwa, line 200: :nit
```suggestion
    <value>The cmdlet cannot run because Authentication is set to {0} and Connect-Graph was not called. Invoke 'Connect-Graph' or specify an authentication token to 'UserProvidedToken' then retry.</value>
```

finsharp: The original version is attempting to inform the user to do the following. 
`Invoke-GraphRequest -Token $token -Authentication UserProvidedToken`

PR #319, opened 2020-08-10
Update OpenAPI docs

peombwa:
This PR updates the current modules mapping config and refreshes the OpenAPI docs to the latest.

The PR also adds:
- Directive to rename /$ref cmdlet to *ByRef e.g.  New-MgGroupOwnerByRef.
- Directive to add new modifier to values properties to avoid conflicts with IAssociativeArray.


openApiDocs/beta/Teams.Chats.yml
```diff
@@ -535,15 +542,104 @@ paths:
...
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: teamsAppInstallation
+      requestBody:
+        description: New navigation property ref values
+        content:
+          application/json:
+            schema:
+              type: String
```
peombwa, line 604: OpenAPI is case sensitive. $ref request bodies should be of type `string` and not `String`.

peombwa: Shouldn't this type be a dictionary? [OpenAPI dictionary](https://swagger.io/docs/specification/data-models/dictionaries/) i.e..
``` yml
      requestBody:
        description: New navigation property ref values
        content:
          application/json:
            schema:
              type: object
              additionalProperties: true
        required: true
```
cc @darrelmiller, @ddyett

PR #352, opened 2020-08-27
Add -CountVariable parameter

### LINKED ISSUE #152: Add better support for -count parameter
Currently, using -count parameter results in an invalid count option.
Repro
Get-MgUserMessage -UserId {user-id} -Count
Get-MgUserMessage_List: 'True' is not a valid count option.
Cause
The service expects the $count value to be lowercase true, yet the generated client sends a pascal case True.
[AB#5873](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5873)

peombwa:
This PR closes #152 by proposing the following changes:
- Ensures all `$count=true` is always in lowercase. MS Graph doesn't support Camel cased count values.
- Adds a runtime `-CountVariable` parameter to all cmdlets that return a collection. This will be used by customers to get back a deserialized value of `@odata.count` returned in the response body. By default, this variable will be stored in the global scope.
- Adds `-CV` alias for `-CountVariable`.
- Makes `-Count` parameter private i.e. it won't be exposed to customers.
- Ensures `$count=true` is always set with `-CountVariable` is present.

### Proposed use
image

### DIFF
--- a/src/readme.graph.md
+++ b/src/readme.graph.md
@@ -469,19 +469,27 @@ directive:
       {
         return $;
       } else {
-        // Add custom -PageSize parameter to *_List cmdlets that support Odata next link.
         let odataNextLinkRegex = /(^\s*)(if\s*\(\s*result.OdataNextLink\s*!=\s*null\s*\))/gmi
         if($.match(odataNextLinkRegex)) {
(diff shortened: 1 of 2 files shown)


finsharp (approved):

finsharp:
@peombwa Ready to Merge?

Issue #381, opened 2020-09-17
Connect-Graph Should Accept Comma Separated String of Scopes


peombwa:
Connect-Graph -Scopes "User.ReadBasic.All, Group.ReadWrite.All, TeamMember.ReadWrite.All" fails with:
![image](https://user-images.githubusercontent.com/7061532/93509727-0000db00-f8d5-11ea-9c20-d32f2ef14e02.png)

Connect-Graph -Scopes "User.ReadBasic.All", "Group.ReadWrite.All", "TeamMember.ReadWrite.All" Passes.
[AB#6081](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6081)

Issue #142, opened 2020-03-17
Connect-Graph Enhancements - Metaissue


darrelmiller:
Related to issues: #125 ,  #123, #102, #46, #52, #113 
[AB#7439](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7439)

tseward:
I would rename `Connect-Graph` to `Connect-MgService`. This makes a bit more sense than `Connect-MgGraph` (Connect Microsoft Graph Graph?) and `Connect-MgAccount` since we're connecting to the Microsoft Graph _service_ and not a reference to an _account_.

Plus it would be like the SPO cmdlets and hopefully would start a movement to get other services to adopt the same format :-)

darrelmiller:
@tseward Thanks for your input.  I shopped the "service" suffix around a bit and the feedback is that the term is a bit overloaded and it conflicts with some of the other work that is going on.   There wasn't much enthusiasm for the Account suffix and there wasn't much objection to `Connect-MgGraph`.  So for the moment, we are planning to go with Connect-MgGraph and alias Connect-Graph so that we don't break existing folks.

darrelmiller:
Parameter | Delegated | Application-Only | MSI | Plan |
-- | -- | -- | --| --|
-Scopes | X |   | |  
-TenantId | X | X |  |
-ClientId (ApplicationId) |   | X |  Alias ApplicationId |
-CertificateName |   | X |  |
-CertificateThumbprint |   | X |  |
-ForceRefresh | X | X |  |
-AccessToken | ? | ? | ?| Post V1 |
-Credential |   | ? |  Post V1 | 
-Environment | X | X | X | Post V1 |
-Identity |   |   | X| Waiting for Graph Core 2 | 
-Certificate |   | ? |  Post V1 |
-ContextName | ? | ? |  No decision | 
-UseDeviceAuthentication | ? |   |   Default in Linux, Option in Windows|

BrianTJackett:
On my work machine I have a number of Connect-* commandlets.  Here are a few:

- Connect-AzureAD              <-- only the module prefix of "AzureAD"
- Connect-SPOService          <-- <prefix> + "Service", already mentioned above
- Connect-AzAccount           <-- <prefix + "Account", not really notion of Account in Graph
- Connect-PnPOnline            <-- <prefix> + "Online", to distinguish from on-prem SharePoint
[...]

PR #384, opened 2020-09-18
Connect and Disconnect-Graph Aliases

peombwa:
This PR proposed the following changes:
1. Renames `Connect-Graph` and `Disconnect-Graph` to `Connect-MgGraph` and `Disconnect-MgGraph` respectively. Their old names are then use as aliases to avoid breaking change. See customer ask https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/142#issuecomment-606776307.
2. Adds support for comma separates scopes. This was previously broken. Closes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/381.

### Open Question
- Should we also alias `Invoke-GraphRequest`? If so, what name should we use.


darrelmiller:
As much as I don't like it, I think we should create the alias `Invoke-MgGraphRequest` just for consistency

darrelmiller (dismissed):

peombwa:
I've renamed `Invoke-GraphRequest` to `Invoke-MgGraphRequest` then aliased it to `Invoke-GraphRequest`.

darrelmiller (approved):

darrelmiller:
Thanks Peter!

PR #397, opened 2020-09-23
Generate profiles in same Stage as Generate and Build Service Modules

finsharp:
Generate profiles in same Stage as Generate and Build Service Modules
Bump to v 1.0.1


peombwa (approved):
LGTM

PR #440, opened 2020-10-22
Allow Connect-Graph to take an x509Certificate.

### LINKED ISSUE #102: Allow Connect-Graph authentication via x509 certificate provided in-memory vs. having to be in local machine store on disk
We have a lot of scenarios where we want to run automations against the Microsoft Graph from Azure Automation, and, unless we curated our own hybrid workers on our own VM's where we have full control over the local cert store, it's often not feasible in this type of scenario to import a cert for authentication into the local cert store. (We still need to test to see if this is possible in a cloud ephemeral worker in Azure Automation -- even if it is, leaving a cert around in seems sort of dicey, hopefully workers are cleared between runs, but you never know...)

In any case, it would be great to be able to pass a certificate to `Connect-Graph` in-memory to facilitate scenarios where you can't reasonably import a cert from the local cert store for authentication. (Or, another alternative would be to allow the use of client_id + client_secret, but there may have been good reasons why client_secret support wasn't implemented; certs may be preferred.) 
[AB#6431](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6431)

finsharp:
Allow Connect-Graph to take an x509Certificate.

Such as:-

```
$password= "SomePassword"
$passwordSecure = ConvertTo-SecureString -AsPlainText -Force $password

$cert = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2("C:\Certificates\mycert.pfx", $passwordSecure)
Connect-Graph -Certificate $cert -TenantId <some-tenant-id> -ClientId <some-client-id>
```
Resolves #102


src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -71,6 +75,27 @@ internal static IAuthenticationProvider GetAuthProvider(IAuthContext authContext
             }
             return authProvider;
         }
+        /// <summary>
+        ///     CGet a
```
peombwa, line 79: Missing comments.

src/Authentication/Authentication.Test/Helpers/AuthenticationHelpersTests.cs
```diff
@@ -89,10 +91,50 @@ public void ShouldUseClientCredentialProviderWhenAppOnlyContextIsProvided()
...
+            // Act
+            IAuthenticationProvider authProvider = AuthenticationHelpers.GetAuthProvider(appOnlyAuthContext);
+
+            // Assert
+            Assert.IsType<ClientCredentialProvider>(authProvider);
+            var clientCredentialProvider = (ClientCredentialProvider) authProvider;
+            // Assert: That the certificate created and set above is the same as used here.
+            Assert.Equal(clientCredentialProvider.ClientApplication.AppConfig.ClientCredentialCertificate, certificate);
+            GraphSession.Reset();
+        }
```
peombwa, line 118: nit: spacing between methods.

peombwa:
@finsharp  What does pipeline `microsoftgraph.msgraph-sdk-powershell (1)` do? It seems to be failing for all PRs. If it's needed, then we should rename it to something useful and figure out why it's failing.

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -71,6 +75,27 @@ internal static IAuthenticationProvider GetAuthProvider(IAuthContext authContext
             }
             return authProvider;
         }
+        /// <summary>
+        ///     Gets a  certificate based on the current context.
```
peombwa, line 79: nit:
```suggestion
        ///     Gets a certificate based on the current context.
```

peombwa (approved):

ddyett:
can we add a sample with this usage once approved?

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -71,6 +75,27 @@ internal static IAuthenticationProvider GetAuthProvider(IAuthContext authContext
...
+            if (!string.IsNullOrWhiteSpace(context.CertificateThumbprint))
+            {
+                return GetCertificateByThumbprint(context.CertificateThumbprint);
+            }
+            
+            else if (!string.IsNullOrWhiteSpace(context.CertificateName))
+            {
+                return GetCertificateByName(context.CertificateName);
+            }
+            else
```
ddyett, line 94: seems like there should be a check for certificate null

src/Authentication/Authentication/Helpers/AuthenticationHelpers.cs
```diff
@@ -71,6 +75,27 @@ internal static IAuthenticationProvider GetAuthProvider(IAuthContext authContext
...
+            {
+                return context.Certificate;
```
ddyett, line 96: if a certificate does exist shouldn't it take precedence over thumprint or name

finsharp: The original two certificate acquisition methods (by Name and Thumbprint) should take precedence since they are what our customers were used to, It would be the least surprising change.

peombwa:
@finsharp The ADO pipelines are failing because we've exhausted our free artifacts storage tier (2 GB). We need to delete old versions of the artifacts then setup a retention policy. See https://docs.microsoft.com/en-us/azure/devops/artifacts/start-using-azure-artifacts?view=azure-devops. Only a feed admin can make these changes.

peombwa (approved):

Issue #479, opened 2020-12-04
Add JSON output support to Invoke-GraphRequest


adamedx:
When building new commands on top of `Invoke-GraphRequest`, it may be necessary to get the response in JSON form (e.g. for interop with other tools that take in JSON). A workaround today is to convert the deserialized output back to JSON (e.g. `ConvertTo-JSON`, but that's obviously inefficient.

Something like a `-Json` option on the command would meet the requirement here. 
[AB#6854](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6854)

Issue #454, opened 2020-11-06
Allow me to get back the raw JSON output from any request easily with a `-Raw` common parameter


KirkMunro:
I would like to be able to invoke any Graph cmdlet with a `-Raw` common parameter, with the results of that invocation returning the raw JSON output from the web request. This would be very helpful because Graph cmdlets allow me to easily write a command in a self-documenting way using PowerShell syntax while allowing me to get back the raw JSON output without it being processed/converted into custom objects for scenarios where I want to pass that JSON off to another service.

Today there are many scenarios where I am invoking Graph cmdlets, which then send receive results in JSON, convert it into objects, which I then convert back into JSON to send off to another service for processing. I'd like to avoid the overhead involved in converting from JSON to objects and then back into JSON, especially since the JSON isn't always converted into objects in a way that I would expect.

I know that I can use Invoke-MgGraphRequest to achieve this; however, that removes the benefit of being able to write self-documenting script logic that leverages PowerShell's strengths. A `-Raw` common parameter to get back just the JSON would give me the best of both worlds when I need it. 
[AB#7373](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7373)

finsharp:
@KirkMunro All the modules return objects that implement a `ToJsonString()`
Such as below :-
```
$users = Get-MgUser -All
$users.ToJsonString()
```

It works for both for collections and single objects. 

cc @peombwa

KirkMunro:
@finsharp: Thank you for sharing that, I had not thought to look for such a method.

Yes, the JSON produced by that method would suit my needs. As long as I can get back the raw JSON from a single graph API, ideally without any overhead involved in converting from JSON to PSObject since I wouldn't need that to be done, I'll be happy.

Also just to call it out, my expectation is that `-Raw` would result in the raw JSON being returned like it would from the Graph REST API invocation. i.e. If multiple records are being looked up and would be returned on a single page as a single JSON string, then just give me that string, but if multiple records are being looked up and the data would be returned from Graph on multiple pages, then stream the JSON strings for each of those pages out of the pipeline, and I'll handle joining the JSON arrays myself.

finsharp:
@KirkMunro We have added this in linked PR, please take a look and see if its satisfactory.

PR #499, opened 2021-01-13
Allow caller to Specify desired OutputType

### LINKED ISSUE #480: Expose full Graph response from `Invoke-GraphRequest`
In order to interop with other tools or perform custom handling of responses from MS Graph, it is useful to get the entire Content of the response in both success and error cases. An option that allows `Invoke-GraphRequest` to simply return the entire response without interpreting it, e.g. a `-Raw` option, would allow custom tooling to fully exploit the protocol and allow for advanced debugging. 
[AB#6855](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6855)

finsharp:
Add `-OutputType [HashTable | PSObject | HttpResponseMessage | Json] ` parameter to enable caller decide preferred format. `HashTable` is the default. 
****PSObject for a Single Entity****
PsObject
****PSObject for a collection****
PsObjectArray
****HttpResponseMessage****
NewHttpReponseMessage

Resolves #480 #479 #454


KirkMunro:
I think you need to simplify the variables that are used here, and fit them more into existing PowerShell paradigms.

For example, instead of:
* returning a custom PowerShell object by default
* returning the HttpResponseMessage if the `-Raw` switch is used
* returning the JSON string if the `-Json` switch is used

Why not provide an `-OutputType` parameter that is an enumeration. The enumeration would have three values: `PSObject`, `Json`, and `HttpResponseMessage`. The default value would be `PSObject`. This would allow the cmdlets to work like they already do today, while allowing users to decide if they want the raw JSON or the HttpResponseMessage object, without having to deal with multiple parameters that are mutually exclusive of one another.

Also, for the `-OriginalRequestClone "createRequestClone"` parameter that creates a "magic" `$requestClone` variable in the current scope, that is not consistent with existing methods used in PowerShell is used to output values into a variable. For example, PowerShell supports common `-OutVariable`, `-ErrorVariable`, `-WarningVariable`, and `-PipelineVariable` parameters that are used to redirect various things into PowerShell variables. The user provides the variable name, and the content is assigned to that variable. The user can choose to overwrite the content of the variable, or add content to the variable. For instance, consider this script:

```powershell
# Get some services, output them to the console, and store them in $myServices
gsv -name wuauserv,bits -ov myServices
# Show the services that were stored
$myServices
# Get some more services, and _add_ them to the myServices variable (note the "+" prefix)
gsv -name k* -ov +myServices
# Show the updated services collection
$myServices
```

This is the model that should be followed for requesting the HTTP request. If you were to apply this to the Graph cmdlets, the command might look like this:

```powershell
# Get users, output them, and capture the HTTP request
Get-MgUser -HttpRequestVariable httpRequest
# Look at the request
$httpRequest
```

In this scenario, you are not dealing with data streams, so I don't think there is a lot of value in supporting the "+" prefix to add another request to an existing variable, other than to maintain consistency with the model that is already in place for capturing content in variables, so it's up to you whether or not you want to add that.

I think these two sets of changes would make for a much better experience for PowerShell scripters that is easier to learn and use.

finsharp:
@KirkMunro Fair points. I'll make changes.

src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
```diff
@@ -420,50 +448,73 @@ private Uri PrepareUri(HttpClient httpClient, Uri uri)
         ///     Process Http Response
         /// </summary>
         /// <param name="response"></param>
-        internal void ProcessResponse(HttpResponseMessage response)
+        internal async Task ProcessResponse(HttpResponseMessage response)
```
peombwa, line 451: Method name should be `ProcessResponseAsync`.

src/Authentication/Authentication/Helpers/WebResponseHelper.cs
```diff
@@ -34,5 +33,93 @@ internal static class WebResponseHelper
...
+                clone.Properties.Add(prop);
+            }
+
+            foreach (var header in original.Headers)
+            {
+                /*
+                **temporarily skip cloning telemetry related headers**
+                clone.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                */
+                if (!"x-ms-unique-id".Equals(header.Key) && !"x-ms-client-request-id".Equals(header.Key) && !"CommandName".Equals(header.Key) && !"FullCommandName".Equals(header.Key) && !"ParameterSetName".Equals(header.Key) && !"User-Agent".Equals(header.Key))
```
peombwa, line 64: Do we really use these request headers in our clients? Here is the complete list of all the headers we send https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/middleware/TelemetryHandler.md#ideal-metadata-capture.

ddyett:
@finsharp do we have some limitation on the azure machines we need to look at.  The errors for the integrated build seem to be indicating that.

src/Authentication/Authentication/Models/OutputType.cs
```diff
@@ -0,0 +1,29 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph.PowerShell.Authentication.Models
+{
+    /// <summary>
+    ///     Output data format to return to the caller
```
ddyett, line 8: NIT: extra space

src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
```diff
@@ -416,60 +425,50 @@ private Uri PrepareUri(HttpClient httpClient, Uri uri)
...
-                    catch (NotSupportedException)
-                    {
-                        encodingVerboseName = encoding.EncodingName;
-                    }
-
-                    // NOTE: Tests use this verbose output to verify the encoding.
-                    WriteVerbose(Resources.ContentEncodingVerboseMessage.FormatCurrentCulture(encodingVerboseName));
-                    WriteObject(str.TryConvertToJson(out var obj, ref ex) ? obj : str);
+                    case OutputType.HashTable:
+                        var hashTable = responseString.ConvertFromJson(true, null, out error);
```
ddyett, line 449: what happens in the case on non-json responses?

src/Authentication/Authentication.Test/Helpers/StringUtilTests.cs
```diff
@@ -1,37 +1,55 @@
...
-        public static IEnumerable<object[]> TestJsonData =>
-            new List<object[]>
-            {
-                new object[] { TestJsonArray},
-                new object[] { TestJsonObject}
-            };
+
+        [Theory]
+        [MemberData(nameof(TestJsonData))]
+        public void ShouldReturnPsObject(string jsonString)
```
ddyett, line 37: this looks like it's testing if we call ConvertFromJson correctly we get the right object but not the actual system.  I think Peter enabled some integration tests so could we use those to hit the real code.

finsharp: Added Pester tests for all scenarios, see `src/Authentication/Authentication/test/Authentication.Tests.ps1`

src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
```diff
@@ -416,60 +425,50 @@ private Uri PrepareUri(HttpClient httpClient, Uri uri)
...
         /// </summary>
         /// <param name="response"></param>
-        internal void ProcessResponse(HttpResponseMessage response)
+        internal async Task ProcessResponseAsync(HttpResponseMessage response)
         {
             if (response == null) throw new ArgumentNullException(nameof(response));
-
-            var baseResponseStream = response.GetResponseStream();
-
+            var responseString = await response.Content.ReadAsStringAsync();
```
peombwa, line 442: This should be within `if (ShouldWriteToPipeline) {` block since it is only ever used when ShouldWriteToPipeline is true.

src/Authentication/Authentication/Helpers/Errors.cs
```diff
@@ -20,5 +20,8 @@ public static class Errors
         public const string InvokeGraphRequestInputFileNoPathResolvedException = nameof(InvokeGraphRequestInputFileNoPathResolvedException);
         public const string InvokeGraphRequestInputFileNotFilePathException = nameof(InvokeGraphRequestInputFileNotFilePathException);
         public const string InvokeGraphRequestMissingAuthenticationContext = nameof(InvokeGraphRequestMissingAuthenticationContext);
+        public const string InvokeGraphRequestEmptyKeyInJsonString = nameof(InvokeGraphRequestEmptyKeyInJsonString);
```
peombwa, line 23: Error constants should ideally go in `ErrorConstants.cs` - https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/features/IGRImprovements/src/Authentication/Authentication/ErrorConstants.cs.

src/Authentication/Authentication/test/Authentication.Tests.ps1
```diff
@@ -0,0 +1,46 @@
...
+         $hashTable = Invoke-GraphRequest -OutputType Hashtable -Uri "https://graph.microsoft.com/beta/users"  | Should -BeOfType [System.Collections.Hashtable]
+    }
+
+    It 'ShouldReturnJsonStringBeta' {
+         $jsonString = Invoke-GraphRequest -OutputType Json -Uri "https://graph.microsoft.com/beta/users"  | Should -BeOfType [System.String]
+    }
+
+    It 'ShouldReturnHttpResponseMessageBeta' {
+        $httpResponseMessage = Invoke-GraphRequest -OutputType HttpResponseMessage -Uri "https://graph.microsoft.com/beta/users" | Should -BeOfType [System.Net.Http.HttpResponseMessage]
+    }
```
peombwa, line 45: Consider adding tests for single entity responses and not just collections to test your deserialization logic. A call to get a user in your tenant should do the trick e.g., `https://graph.microsoft.com/v1.0/users/$env.DefaultUserId`.

src/Authentication/Authentication/test/Authentication.Tests.ps1
```diff
@@ -0,0 +1,46 @@
...
+        $loadEnvPath = Join-Path $PSScriptRoot '..\loadEnv.ps1'
+    }
+    . ($loadEnvPath)
+    $ModuleName = "Microsoft.Graph.Authentication"
+    $ModulePath = Join-Path $PSScriptRoot "..\$ModuleName.psd1"
+    Import-Module $ModulePath -Force
+}
+Describe 'Invoke-GraphRequest' {
+    BeforeAll {
+         Connect-Graph
```
peombwa, line 13:
```suggestion
         Connect-MgGraph
```
nit: Let's avoid using aliases where possible since they could change.

src/Authentication/Authentication/test/loadEnv.ps1
```diff
@@ -0,0 +1,29 @@
...
+}
+$env = @{}
+if (Test-Path -Path $envFilePath) {
+    $env = Get-Content (Join-Path $PSScriptRoot $envFile) | ConvertFrom-Json
+} else {
+    $env.TenantIdentifier = ${env:TENANTIDENTIFIER}
+    $env.ClientIdentifier = ${env:CLIENTIDENTIFIER}
+    $env.CertificateThumbprint = ${env:CERTIFICATETHUMBPRINT}
+}
+$PSDefaultParameterValues=@{"Connect-Graph:TenantId"=$env.TenantIdentifier; "Connect-Graph:ClientId"=$env.ClientIdentifier; "Connect-Graph:CertificateThumbprint"=$env.CertificateThumbprint}
```
peombwa, line 29:
```suggestion
$PSDefaultParameterValues=@{"Connect-MgGraph:TenantId"=$env.TenantIdentifier; "Connect-MgGraph:ClientId"=$env.ClientIdentifier; "Connect-MgGraph:CertificateThumbprint"=$env.CertificateThumbprint}
```
nit: Let's avoid using aliases where possible since they could change.

src/Authentication/Authentication/test/utils.ps1
```diff
@@ -0,0 +1,24 @@
...
+    } else {
+        return -join ((48..57) + (97..122) | Get-Random -Count $len | % {[char]$_})
+    }
+}
+$env = @{}
+function setupEnv() {
+    # Preload subscriptionId and tenant from context, which will be used in test
+    # as default. You could change them if needed.
+    $env.SubscriptionId = (Get-AzContext).Subscription.Id
+    $env.Tenant = (Get-AzContext).Tenant.Id
```
peombwa, line 13: Do we use these anywhere? If not, then we can remove them.

src/Authentication/Authentication/Properties/Resources.resx
```diff
@@ -199,10 +199,19 @@
   <data name="NotConnectedToGraphException" xml:space="preserve">
     <value>The cmdlet cannot run because Authentication is set to {0} and Connect-Graph was not called. Invoke 'Connect-Graph' or specify Authentication to be 'UserProvidedToken' and Provide a Token then retry</value>
```
peombwa, line 200:
```suggestion
    <value>The cmdlet cannot run because Authentication is set to {0} and Connect-MgGraph was not called. Invoke 'Connect-MgGraph ' or specify Authentication to be 'UserProvidedToken' and Provide a Token then retry</value>
```
nit: Let's avoid using aliases where possible since they could change.

src/Authentication/Authentication/Properties/Resources.resx
```diff
@@ -199,10 +199,19 @@
...
   </data>
+  <data name="MissingAuthenticationContext" xml:space="preserve">
+    <value>The cmdlet cannot run because {0} is missing and 'Authentication' is set to 'Default'. Please call 'Connect-Graph' then try again.</value>
```
peombwa, line 203:
```suggestion
    <value>The cmdlet cannot run because {0} is missing and 'Authentication' is set to 'Default'. Please call 'Connect-MgGraph' then try again.</value>
```
nit: Let's avoid using aliases where possible since they could change.

src/Authentication/Authentication/Microsoft.Graph.Authentication.csproj
```diff
@@ -22,7 +22,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Graph.Auth" Version="1.0.0-preview.5" />
-    <PackageReference Include="Microsoft.Graph.Core" Version="1.20.1" />
+    <PackageReference Include="Microsoft.Graph.Core" Version="1.24.0" />
```
peombwa, line 25:
```suggestion
    <PackageReference Include="Microsoft.Graph.Core" Version="1.20.1 />
```
Bumping to 1.24.0 without setting the correct dependency versions results in missing dll in PS 5.1. 
image

Consider reverting back to 1.20.1 since I'll be updating this lib when I enable CAE.

finsharp: There is a bug fix in 1.24, regarding header usage that went out last year https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/123.
We need the PR to be able to know what the response type is.

peombwa: Then we should update the dependencies to avoid the error above in PS 5.1.

peombwa (approved):

PR #528, opened 2021-01-28
Remove *-UserAuthentication commands

andres-canello:
Remove *-UserAuthenticationMethod commands from the Identity.SignIns module.


src/Identity.SignIns/Identity.SignIns/readme.md
```diff
@@ -48,6 +48,10 @@ directive:
       subject: ^(DataPolicyOperation)(\1)+
     set:
       subject: $1
+# Remove cmdlets
+  - where:
+      subject: ^UserAuthenticationMethod$
```
peombwa, line 53:
```suggestion
      subject: ^UserAuthentication$
```
My bad, the subject regex should be `^UserAuthentication$`.

peombwa (approved):

ddyett:
@peombwa is there an issue for the reason this needed to be removed?

andres-canello:
> @peombwa is there an issue for the reason this needed to be removed?

@ddyett We don't have any methods associated to the property "Authentication", this is why I'm asking to remove the associated commands.

peombwa:
@ddyett `Get-MgUserAuthentication` currently calls `https://graph.microsoft.com/beta/users/{user-id}/authentication` which is an invalid path and return a 404. The commands should be `*-MgUserAuthentication*`.

PR #540, opened 2021-02-02
Cert Thumbprint Validation & Parameter Classification

### LINKED ISSUE #226: Connect-MicrosoftGraph invalid CertificateThumbprint still connects
When calling the Connect-Graph cmdlet, if I pass in an invalid certificatethumbprint I am still able to connect and interact with the APIs.

For example, let's say that my certificate thumbprint is "22E75E5D37791759FE89ED48FF843BB5132B8E44". I am able to connect passing in "22E75E5D37791759FE89ED48FF843BB5132B8E44s" with an extra 's' at the end (which was a typo to begin with). Even if the certificate thumbprint is invalid it still let me connect. Is there internal logic in the connect-graph command that truncates the value down to 40 characters no matter what? In my opinion, it should fail to connect me if I do a typo in the certificate thumbprint even if the first 40 characters are valid. 
[AB#7404](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7404)

### LINKED ISSUE #510: Connect-MgGraph: -AccessToken ParameterSet Issues (dead parameters, silent failures, etc.)
Issues found while developing https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/509

- [ ] `-TenantId` parameter does nothing, it still just uses the tenant specified in the JWT accesstoken
- [ ] Should do a simple test such as `/me` to verify the token is valid and immediately throw error if not. Doesn't seem to error until attempted to be used once.
- [ ] Silently fails if token is invalid, doesn't even try to check it according to fiddler trace, should test and throw error.
- [ ] `-ForceRefresh` does nothing (it can't because no refresh token is included) and should be removed from parameterset
- [ ] `-Environment` does nothing 
- [ ] `-ContextScope CurrentUser` does not seem to work, it doesn't preserve for the current user, starting a new process bails on it
[AB#7711](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7711)

peombwa:
This PR:
- Closes #226 by validating the length of the certificate thumbprint set by a customer. An x.509 certificate thumbprint is expected to have a length of 40 since its a SHA-1 hash of the entire certificate. A SHA-1 function will always produce a 160-bit (20-byte = 40-hexadecimal) hash value.
Reference:
  - https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.thumbprint?view=net-5.0#remarks.
  - https://stackoverflow.com/questions/26441931/sha1-encoding-to-hex-has-40-characters-and-160-bits
- Closes #510 by removing `TenantId`, `ForceRefresh`, and `ContextScope` from `-AccessToken` [parameter set](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/cmdlet-parameter-sets?view=powershell-7.1#:~:text=PowerShell%20uses%20parameter%20sets%20to%20enable%20you%20to,specified%20by%20the%20user.%20Examples%20of%20parameter%20sets). This leaves us with a clean and accurate Get-Help for `Connect-MgGraph`:
image


finsharp (approved):

gaikovoi:
@peombwa 
What is the reasone to remove ContextScope from  AccessTokenParameterSet?

#572

PR #576, opened 2021-03-09
Add build agents parameter

finsharp:
Add Build Agents Parameter to allow selecting build agent during runs.

### DIFF
--- a/.azure-pipelines/common-templates/download-openapi-docs.yml
+++ b/.azure-pipelines/common-templates/download-openapi-docs.yml
@@ -8,11 +8,14 @@ parameters:
   - name: BaseBranch
     type: string
     default: "dev"
+  - name: BuildAgent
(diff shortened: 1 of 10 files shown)


.azure-pipelines/common-templates/download-openapi-docs.yml
```diff
@@ -8,11 +8,14 @@ parameters:
   - name: BaseBranch
     type: string
     default: "dev"
+  - name: BuildAgent
+    default: MsGraphBuildAgentsLinux
```
peombwa, line 12: Can the SDK be built on Linux?

finsharp: It cant because of the net461 Authentication Tests. I reverted the change to use the new updated agent.

peombwa (approved):

Issue #555, opened 2021-02-12
Implement interactive sign-in


darrelmiller:
Device code flow is blocked by many corporations that have Intune managed devices.  The reason they block Device Code flow is because it is possible to run an app on an unmanaged device, but then do the device code flow handshake on a managed device, therefore circumventing the conditional access policy.

It is my understanding that MSAL.NET supports interactive flow cross platform, so we should investigate enabling this so that users can use our PowerShell modules in these Intune Managed environments. 
[AB#8122](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8122)

finsharp:
Depends on #534

PR #517, opened 2021-01-25
Add CAE Support

peombwa:
This PR adds Continuous Access Evaluation (CAE) support to SDK by:
- Updating MSAL, MS Graph .NET core, and MS Graph Auth libs to versions that support CAE.
- Adding `cp1` client capability to our public client builder. This capability sends an` xms_cc` param to `EvoSTS` which tells it that the client is capable of handling CAE scenarios and should be issues with CAE tokens.

#### Claims in JWT:
image

#### Reference:
- [What is CAE](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/concept-continuous-access-evaluation) 
- [How to enable CAE](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-resilience-continuous-access-evaluation#the-code)

Closes #508.

Issue #415, opened 2020-10-01
Consider adding interactive login for framework usage?


mlhickey:
Ever consider including Interactive for use in a framework world, e.g.
ConnectGraph.cs
#if NETFRAMEWORK
                    clientApplication = (authProvider as InteractiveAuthenticationProvider).ClientApplication;
#else
                    clientApplication = (authProvider as DeviceCodeProvider).ClientApplication;
#endif

AuthenticationHelper.cs
#if NETFRAMEWORK
                return new InteractiveAuthenticationProvider(publicClientApp, authContext.Scopes);
#else
                return new Microsoft.Graph.Auth.DeviceCodeProvider(publicClientApp, authContext.Scopes, async (result) =>
                {
                    await Console.Out.WriteLineAsync(result.Message);
                });
#endif 
[AB#7398](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7398)

MIchaelMainer:
This will require us to follow up with MSAL.

peombwa:
Related to #334

peombwa:
This has been resolved by #618 and will be available in v1.6.0.

Issue #334, opened 2020-08-12
Change default sign-in behaviour to use Interactive Browser Sign-in


darrelmiller:
This will be the default behaviour on Windows.  Linux/MacOS will use Device Code flow by default.
We need to add a parameter to enable device code flow on windows. See #142
[AB#5758](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5758)

peombwa:
Waiting on https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1991

finsharp:
Depends on #534

peombwa:
This has been resolved by #618.

PR #585, opened 2021-03-13
Safe Dependency Resolution For Side-By-Side Loading With Other Modules

### LINKED ISSUE #534: Implement assembly resolve handler for side-by-side loading with other modules
We should implement a more robust assembly resolve handler for side-by-side loading with other modules that use the same dependencies as us.

This implementation should work in both PS 5.1 and PS 6+.

Related to #532  
[AB#7889](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7889)

peombwa:
This PR closes #534 by implementing an [assembly resolve handler](https://docs.microsoft.com/en-us/dotnet/standard/assembly/resolve-loads) that handles loading our dependencies for side-by-side loading with other modules. This change involves:
- Dividing auth module into two assemblies: `Microsoft.Graph.Authentication.Core` and `Microsoft.Graph.Authentication`.
image
`Microsoft.Graph.Authentication.Core` and `Microsoft.Graph.Authentication` will be loaded by PowerShell, and all our dependencies will be loaded by `Microsoft.Graph.Authentication.Core` into the execution context.
  - `Microsoft.Graph.Authentication.Core` specifies and makes calls to our dependencies e.g. MSAL.
  - `Microsoft.Graph.Authentication` contains implementations of our cmdlets and calls methods in `Microsoft.Graph.Authentication.Core`.
- Updates ADO pipeline scripts to build, sign, and pack Microsoft.Graph.Authentication and Microsoft.Graph.Authentication.Core assemblies into our Authentication module.
- Adds pester tests to validate changes and ensure no breaking change is introduced.
- Removes unused ADO pipeline scripts from `.\azure-pipelines`.

This PR should unblock #555, #415, #334, and #517


finsharp:
Can this PR lead us to a path of building and testing on linux?

src/Authentication/Authentication/Utilities/DependencyAssemblyResolver.cs
```diff
@@ -0,0 +1,101 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Reflection;
+    public static class DependencyAssemblyResolver
+    {
+        // Catalog our dependencies here to ensure we don't load anything else.
+        private static IReadOnlyDictionary<string, Version> Dependencies = new Dictionary<string, Version>
+        {
+            { "Microsoft.Identity.Client", new Version("4.23.0.0") },
```
finsharp, line 16: Can the resolution logic use a file instead of being explicitly set here?

For example read the csproj to determine this list?

peombwa: We can't access csproj here since this is done at runtime and not build time. We need to know which assembly and version to load on our own when PowerShell imports a module thus the need for an in-memory dictionary.

peombwa:
> Can this PR lead us to a path of building and testing on linux?

@finsharp, the scope of this PR is only limited to addressing dependency resolution at runtime. To build on Linux, we will need to overhaul our generation scripts.

finsharp (approved):

PR #599, opened 2021-03-24
Interactive Authentication

### LINKED ISSUE #555: Implement interactive sign-in

finsharp:
Inform the user to switch to Device Code Authentication when Browser and thus interactive auth is not availoable such as on CloudShell.

Allow the user to specify `UseDeviceCode` explicitly.

Correct minor typos

**Examples**
`Connect-MgGraph` 
OS: Windows
**PowerShell** - Opens a Browser Windows/Tab
**Windows PowerShell** - Opens a Pop-Up Window
**Linux (WSL and other no Browser Environments)** - Fails with Instructions to the user on what to do.
image
***Requirements***
Add a Reply/Redirect URI to PowerShell SDK App Registration.

Resolves #555, #334


src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
```diff
@@ -201,15 +220,26 @@ private async Task ProcessRecordAsync()
...
                     GraphSession.Instance.AuthContext = await Authenticator.AuthenticateAsync(authContext, ForceRefresh, _cancellationTokenSource.Token);
                 }
-                catch(Exception ex)
+                catch (Exception ex)
                 {
+                    if (IsUnableToOpenWebPageError(ex))
+                    {
+                        WriteWarning(Resources.InteractiveAuthNotSupported);
+                        WriteDebug(ex.ToString());
+                    }
```
peombwa, line 229: Should we just warn customers that interactive auth is not supported and switch to device code without throwing an exception?

The current design would result in a breaking change for customers who currently use our SDK in environments where browsers are not available i.e. CloudShell, Linux servers, Azure functions e.t.c.

peombwa:
What is/are the redirect URI that should be added to the PowerShell SDK app registration?

finsharp:
> What is/are the redirect URI that should be added to the PowerShell SDK app registration?

@darrelmiller added `http://localhost` to our registration. 
As described here  https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/MSAL.NET-uses-web-browser

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
finsharp: Taking inspiration from [Azure](https://github.com/Azure/azure-powershell/blob/22119cadba08bba03c5f596f92aa7e9bbd7fa05e/src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs#L441-L464), where they inform the customer that Interactive is not supported and requiring them to make the conscious choice to switch to device code.

peombwa: We may have to take a different tangent with this to avoid breaking our customers' scripts. The current approach will force customers running the SDK in non-browser environments to update their scripts and add `-UseDeviceAuthentication`. This is fine if we are also bumping the major version to signify a breaking change; otherwise, we should fall back to the current behavior if we can't open a browser.

@ddyett, @darrelmiller, thoughts on this?

finsharp: image
Shown above, Warning to the user about the fallback.

openApiDocs/beta/Bookings.yml
```diff
@@ -2515,10 +2515,10 @@ components:
       properties:
```
peombwa, line 2515: Could you undo the last commit that added the Open API descriptions? These descriptions are old and unrelated to the PR.

PR #601, opened 2021-03-30
Fix exception thrown when Debug is called for Service modules.

finsharp:
Show full request body and response when Debug is called.


peombwa:
`"The stream was already consumed.."` exception is still thrown when calling commands that make POST calls such as ` Test-MgDirectoryObjectProperty`:

image

peombwa:
We should also honor [$DebugPreference](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#debugpreference) that is set by a customer.
Setting `$DebugPreference` to `continue` only prints out a limited number of debug messages. See repro steps below :
image

peombwa:
It would be nice to add Pester tests for some of these edge cases. e.g.
``` powershell
$ps = [powershell]::Create()
$ps.AddScript(@'
$DebugPreference = 'Continue'

Import-Module .\src\Authentication\Authentication\artifacts\Microsoft.Graph.Authentication.psd1
Import-Module .\src\DirectoryObjects\DirectoryObjects\Microsoft.Graph.DirectoryObjects.psd1

Connect-MgGraph
Test-MgDirectoryObjectProperty -DisplayName "New Name" -EntityType "Group" 
'@).Invoke()

# Assert that debug stream contains the expected debug log and no exception.
if ($ps.Streams.Debug -notLike "*Exception*") { pass test ... }
if ($ps.Streams.Debug -like "*HTTP/1.1 200 OK*) { pass test ...}
....
```

peombwa:
> We should also honor [$DebugPreference](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#debugpreference) that is set by a customer.
> Setting `$DebugPreference` to `continue` only prints out a limited number of debug messages. See repro steps below :
> image

Is there a way for us to honor `$DebugPreference`? Without it, customers running cmdlets in scripts won't be able to bypass the confirm dialog.
image

finsharp:
> > We should also honor [$DebugPreference](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#debugpreference) that is set by a customer.
> > Setting `$DebugPreference` to `continue` only prints out a limited number of debug messages. See repro steps below :
> > image
> 
> Is there a way for us to honor `$DebugPreference`? Without it, customers running cmdlets in scripts won't be able to bypass the confirm dialog.
> image

@peombwa This seems to be a [known issue](https://github.com/PowerShell/PowerShell/issues/4568), especially since there is nothing special we are doing that would disregard the customer preference.

peombwa (approved):

PR #618, opened 2021-04-13
Interactive Authentication

### LINKED ISSUE #555: Implement interactive sign-in

georgend:
- Set InteractiveAuthenticationProvider as the default auth provider. 
- Inform the user of fallback to device code when Interactive Auth is unavailable due to browser not being available such as Azure CloudShell and WSL.
- Minor cleanup, remove unused methods.
- Add default redirect Url to PowerShell SDK App Registration. Done by @darrelmiller as detailed [in MSAL Docs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/MSAL.NET-uses-web-browser)

***Examples***
Environment: **Windows**
PowerShell - Opens a Browser Windows/Tab
Windows PowerShell - Opens a Pop-Up Window
Environment : **WSL**
Falls Back to **DeviceCode**
image
Specifying -UseDeviceAuthentication
image

Resolves #555, #334


darrelmiller:
Hey @finsharp ! It is great to see this work.  I look forward to trying it out.

src/Authentication/Authentication/Properties/Resources.resx
```diff
@@ -223,4 +223,10 @@
   <data name="NonJsonResponseWithoutOutputFilePath" xml:space="preserve">
     <value>Request returned Non-Json response of {0}, Please specify '-OutputFilePath'</value>
   </data>
+  <data name="InteractiveAuthNotSupported" xml:space="preserve">
+    <value>Interactive authentication is not supported in this session, please run cmdlet 'Connect-MgGraph -UseDeviceAuthentication'.</value>
+  </data>
+  <data name="DeviceCodeFallback" xml:space="preserve">
+    <value>Interactive authentication is not supported in this session, Falling Back to DeviceCode. Future versions will not automatically fallback to DeviceCode.</value>
```
peombwa, line 230: nit: 
```suggestion
    <value>Interactive authentication is not supported in this session, falling back to DeviceCode. Future versions will not automatically fallback to DeviceCode.</value>
```

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
```diff
@@ -196,20 +209,46 @@ private async Task ProcessRecordAsync()
...
+                        {
+                            GraphSession.Instance.AuthContext = context;
+                            //Fallback was due to an Exception, ie Browser Could not be Opened.
+                            if (authError.AuthErrorType == AuthErrorType.FallBack)
+                            {
+                                WriteWarning(Resources.DeviceCodeFallback);
+                            }
+                            WriteObject("Welcome To Microsoft Graph!");
+                            break;
+                        }
```
peombwa, line 236: Isn't this a duplicate of line [219](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/618/files#diff-6765c82282b378bf2f4edc2b785424abe653b006678a02f6b9f4cc0317036bc0R219)?

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
```diff
@@ -170,9 +182,10 @@ private async Task ProcessRecordAsync()
                             _cancellationTokenSource.CancelAfter(authTimeout);
                             authContext.AuthType = AuthenticationType.Delegated;
                             string[] processedScopes = ProcessScopes(Scopes);
-                            authContext.Scopes = processedScopes.Length == 0 ? new string[] { "User.Read" } : processedScopes;
+                            authContext.Scopes = processedScopes.Length == 0 ? new[] { "User.Read" } : processedScopes;
                             // Default to CurrentUser but allow the customer to change this via `ContextScope` param.
                             authContext.ContextScope = this.IsParameterBound(nameof(ContextScope)) ? ContextScope : ContextScope.CurrentUser;
+                            authContext.UseDeviceAuth = UseDeviceAuthentication;
```
peombwa, line 188: Consider using an enum instead. e.g:
``` powershell
# UserParameterSet
authContext.AuthProviderType = UseDeviceAuthentication ? AuthProviderType.DeviceCodeProvider : AuthProviderType.InteractiveAuthenticationProvider;
# AppParameterSet
authContext.AuthProviderType = AuthProviderType.ClientCredentialProvider;
# AccessTokenParameterSet
authContext.AuthProviderType = AuthProviderType.None;
```
This will make it easier to retreive the right clinetApplication in AuthenticateAsync.

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
```diff
@@ -196,20 +209,46 @@ private async Task ProcessRecordAsync()
...
+                   type =>
+                   {
+                       // Early Detection of FallBack to DeviceCode
+                       if (type == AuthProviderType.DeviceCodeProviderFallBack)
+                       {
+                           WriteWarning(Resources.DeviceCodeFallback);
+                       }
+                   });
+
+                switch (authError.AuthErrorType)
```
peombwa, line 223: Why not catch these in a try-catch? We are no longer relying on authError to fallback to device code.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -66,45 +66,52 @@ public static IAuthenticationProvider GetAuthProvider(IAuthContext authContext)
...
                         .Create(authContext.ClientId)
                         .WithTenantId(authContext.TenantId)
                         .WithAuthority(authorityUrl)
                         .WithClientCapabilities(new[] { "cp1" })
+                        .WithDefaultRedirectUri()
                         .Build();
-
-                        ConfigureTokenCache(publicClientApp.UserTokenCache, authContext);
+                    ConfigureTokenCache(publicClientApp.UserTokenCache, authContext);
+                    if (authContext.UseDeviceAuth)
```
peombwa, line 80: If you switch to enum, this will be:
```suggestion
                    if (authContext.AuthProviderType == AuthProviderType.DeviceCodeProvider)
```

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -77,36 +106,84 @@ public static async Task<IAuthContext> AuthenticateAsync(IAuthContext authContex
...
 
         /// <summary>
-        /// Signs out of the provided <see cref="IAuthContext"/>.
+        ///  Gets a static instance of IAuthenticationProvider when the client app hasn't changed.
         /// </summary>
-        /// <param name="authContext">The <see cref="IAuthContext"/> to sign-out from.</param>
-        public static void LogOut(IAuthContext authContext)
+        /// <param name="authContext"></param>
+        /// <returns></returns>
+        private static (IAuthenticationProvider authProvider, IClientApplicationBase clientApplication, AuthProviderType authProviderType) GetClientApplication(IAuthContext authContext)
```
peombwa, line 147: This is not needed if `fallBackWarning` is made nullable in `AuthenticateAsync`. See my suggestion above.

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -29,23 +30,51 @@ public static class Authenticator
...
+
+        /// <summary>
+        /// Authenticates the client using the provided <see cref="IAuthContext"/>.
+        /// </summary>
+        /// <param name="authContext">The <see cref="IAuthContext"/> to authenticate.</param>
+        /// <param name="forceRefresh">Whether or not to force refresh a token if one exists.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <param name="fallBackWarning"></param>
+        /// <returns></returns>
+        public static Task<(IAuthContext context, AuthError authError)> AuthenticateAsync(IAuthContext authContext, bool forceRefresh, CancellationToken cancellationToken, Action<AuthProviderType> fallBackWarning)
```
peombwa, line 48: This is not needed if `fallBackWarning` is made nullable in `AuthenticateAsync`. See my suggestion below.

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -29,23 +30,51 @@ public static class Authenticator
         /// <param name="forceRefresh">Whether or not to force refresh a token if one exists.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns></returns>
-        public static async Task<IAuthContext> AuthenticateAsync(IAuthContext authContext, bool forceRefresh, CancellationToken cancellationToken)
+        public static Task<(IAuthContext context, AuthError authError)> AuthenticateAsync(IAuthContext authContext, bool forceRefresh, CancellationToken cancellationToken)
```
peombwa, line 33: This is not needed if `fallBackWarning` is made nullable in `AuthenticateAsync`. See my suggestion below.

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -29,23 +30,51 @@ public static class Authenticator
...
+        /// <summary>
+        /// Authenticates the client using the provided <see cref="IAuthContext"/>.
+        /// </summary>
+        /// <param name="clientApplication"></param>
+        /// <param name="authProvider"></param>
+        /// <param name="authContext"></param>
+        /// <param name="forceRefresh"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        private static async Task<(IAuthContext context, AuthError authError)> AuthenticateAsync(IClientApplicationBase clientApplication, IAuthenticationProvider authProvider, IAuthContext authContext, bool forceRefresh, CancellationToken cancellationToken)
```
peombwa, line 73: Making `fallBackWarning` nullable here will greatly simplify this class.
```suggestion
        private static async Task<(IAuthContext context, AuthError authError)> AuthenticateAsync(IAuthContext authContext, bool forceRefresh, CancellationToken cancellationToken, Action<AuthProviderType> fallBackWarning = null)
```

peombwa (changes requested):

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -29,23 +30,51 @@ public static class Authenticator
...
         {
             try
             {
```
peombwa, line 76: With an optional `fallBackWarning` and `authContext.AuthProviderType` set to an enum, you can now get the clientApplication like this:
```powershell
var authProvider = AuthenticationHelpers.GetAuthProvider(authContext);
IClientApplicationBase clientApplication = null;
switch (authContext.AuthProviderType)
{
    case AuthProviderType.InteractiveAuthenticationProvider:
        clientApplication = (authProvider as InteractiveAuthenticationProvider).ClientApplication;
        if (!(clientApplication as IPublicClientApplication).IsUserInteractive())
        {
            authContext.AuthProviderType = AuthProviderType.DeviceCodeProviderFallBack;
            if (fallBackWarning != null)
            {
                fallBackWarning(authContext.AuthProviderType);
            }
        }
        break;
    case AuthProviderType.DeviceCodeProvider:
        clientApplication = (authProvider as DeviceCodeProvider).ClientApplication;
        break;
    case AuthProviderType.ClientCredentialProvider:
        clientApplication = (authProvider as ClientCredentialProvider).ClientApplication;
        break;
}
```

finsharp:
@peombwa 
There are two scenarios. 
- A user may not be in an interactive session. - We detect this by the call to `IsUserInteractive()`
- A User may be in an interactive session but for some reason (Browser not installed for example) the browser fails to open.  - This can only be known through the exception thrown. Thus we need to fallback to DeviceCode auth in this scenario. Thats where the recursion in the catch block comes in.

src/Authentication/Authentication.Core/Authenticator.cs
```diff
@@ -77,36 +106,84 @@ public static async Task<IAuthContext> AuthenticateAsync(IAuthContext authContex
...
                 JwtHelpers.DecodeJWT(httpRequestMessage.Headers.Authorization?.Parameter, account, ref authContext);
-                return authContext;
+                return (authContext, new AuthError(AuthErrorType.None, null));
             }
             catch (AuthenticationException authEx)
             {
-                if ((authEx.InnerException is TaskCanceledException) && cancellationToken.IsCancellationRequested)
+                //Interactive Authentication Failure: Could Not Open Browser, fallback to DeviceAuth
+                if (IsUnableToOpenWebPageError(authEx))
+                {
```
peombwa, line 115: > There are two scenarios.
> 
> * A user may not be in an interactive session. - We detect this by the call to `IsUserInteractive()`
> * A User may be in an interactive session but for some reason (Browser not installed for example) the browser fails to open.  - This can only be known through the exception thrown. Thus we need to fallback to DeviceCode auth in this scenario. Thats where the recursion in the catch block comes in.

@finsharp, both scenarios can be handled in this method once the [proposed method signature](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/618/files/8e7f5bb2879a9403c6d3ac115e7c59c422e95103#r616269503) is implemented. You can let `fallBackWarning` only log the warning message [without checking the type](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/618/files/8e7f5bb2879a9403c6d3ac115e7c59c422e95103#diff-6765c82282b378bf2f4edc2b785424abe653b006678a02f6b9f4cc0317036bc0R217), and let the caller check the type before invoking the callback action. e.g.
```suggestion
                {
                    if (fallBackWarning != null)
                        fallBackWarning();
```

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
finsharp: The question becomes for predictable scenarios, why throw an exception instead of handling the scenario?

peombwa (approved):

PR #627, opened 2021-04-16
IG: add parameter sets to construct assignment requests

markwahl-msft:
In the Azure AD Identity.Governance module, this customization adds two parameter sets to New-MgEntitlementManagementAccessPackageAssignmentRequest, to simplify creating an Entitlement Management access package assignment request to administratively add or remove access.

For example, if an admin has the ID of an access package in $apid and the ID of a user in $tid, and the AP has a single policy for admin assignment, then to assign the user to that access package with that policy, 
```
$ap = Get-MgEntitlementManagementAccessPackage -AccessPackageId $apid -ExpandProperty accesspackageassignmentpolicies
$polid = $ap.AccessPackageAssignmentPolicies[0].Id
New-MgEntitlementManagementAccessPackageAssignmentRequest -AccessPackageId $apid -AssignmentPolicyId $polid -TargetId $tid
```


peombwa (approved):

src/Identity.Governance/Identity.Governance/custom/New-MgEntitlementManagementAccessPackageAssignmentRequest.ps1
```diff
@@ -0,0 +1,246 @@
...
+      $PSBoundParameters['RequestType'] = $RequestType
+  
+      try {
+        $outBuffer = $null
+        if ($PSBoundParameters.TryGetValue('OutBuffer', [ref]$outBuffer)) {
+            $PSBoundParameters['OutBuffer'] = 1
+        }
+        $subecmd = 'Microsoft.Graph.Identity.Governance.private\New-MgEntitlementManagementAccessPackageAssignmentRequest_CreateExpanded'
+        
+        $subWrappedCmd = $ExecutionContext.InvokeCommand.GetCommand(($subecmd), [System.Management.Automation.CommandTypes]::Cmdlet)
```
peombwa, line 221: See my comments here https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/632.

peombwa, line 221: See my comments here https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/632.

peombwa (changes requested):

src/Identity.Governance/Identity.Governance/custom/New-MgEntitlementManagementAccessPackageAssignmentRequest.ps1
markwahl-msft: Updated to move to process section.

peombwa (approved):
:shipit:

adamedx:
Interesting -- so @markwahl-msft, @peombwa  we could generalize this as a customization approach: if there is an existing command that is functionally correct in that it maps to a single API, we can create a PowerShell-based wrapper. One key question here: is this purely additive, or does it break the existing interface of the command?

@peombwa , is there any reason we can't add pipelining this way? I think this should be feasible.

adamedx:
Also, @markwahl-msft do you plan on making this the default parameterset? And the docs must be provided via the comments of the PowerShell function?

markwahl-msft:
@adamedx Yes this is intended to be the default parameter set.

> And the docs must be provided via the comments of the PowerShell function?

I had tried setting the synopsis and description in the top of the cmdlet ps1 file, but in get-help ... -detailed I'm not seeing the values set there.  Ideally I'd like the help to be able to automatically incorporate content that's already been written in articles such as https://docs.microsoft.com/graph/api/accesspackageassignmentrequest-post for each of the cmdlets that corresponds to a single Graph API call, since those articles already have comprehensive descriptions, examples etc.

peombwa:
@adamedx This is purely additive and does not break existing parameterSets of the same command.

> Is there any reason we can't add pipelining this way? I think this should be feasible.

We would have to apply a wrapper for all generated commands to support pipeline this way. AutoREST has its own piping mechanism which uses an `InputObject`. There is currently a bug with it where it assumes all id parameters are named `Id` and not `{Entity}Id`, like those in our OpenAPI files. e.g. `Get-MgEntitlementManagementAccessPackage -Id` vs `Get-MgEntitlementManagementAccessPackage -AccessPackageId`.

PR #644, opened 2021-04-27
identity governance: entitlement management bulk direct assignment creation

markwahl-msft:
A popular approach used by customers is to bulk create assignment requests for users who do not already have assignments.  Including either explicit named users in the directory, or members of a group.  In future could also include users who are not yet in the tenant.


adamedx:
So to clarify, this is an entirely new command, and not just a new parameterset, correct?

peombwa (approved):

peombwa:
> So to clarify, this is an entirely new command, and not just a new parameterset, correct?

@adamedx, that's right. This is a new command that adds a [functionality that doesn't exist](https://docs.microsoft.com/en-us/graph/api/resources/accesspackageassignment?view=graph-rest-beta#methods) as an API unless you make nested calls.

PR #668, opened 2021-05-19
Enable Support for PowerShell ISE by handling output via Cmdlet methods.

georgend:
On module importation, pass along PSCmdlet to gain access to output methods from the cmdlet.
Remove references to `Console.Out` which is not available in PowerShell ISE. It uses the methods from the Cmdlets to write to outputs correctly. 
**Current Behavior**
ISE Failures
**Proposed Behavior**
new behavior


src/Authentication/Authentication/Common/Output.cs
```diff
@@ -0,0 +1,17 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph.PowerShell.Authentication.Common
+{
+    using System;
+
+    public class Output : IOutput
```
peombwa, line 9: This class should be under `models` directory for consistency - https://github.com/microsoftgraph/msgraph-sdk-powershell/tree/dev/src/Authentication/Authentication/Models.

Also, we should give this class a different name since Output is too generic and can be confused with [OutputType.cs](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Authentication/Authentication/Models/OutputType.cs). Something like `PSOutputStream` or `OutputStream` would be ideal.

src/Authentication/Authentication.Core/Interfaces/IAuthContext.cs
```diff
@@ -26,6 +28,16 @@ public enum AuthProviderType
         ClientCredentialProvider,
         UserProvidedToken
     }
+
+    public interface IOutput
```
peombwa, line 32: Let's move this interface to its own file.

src/Authentication/Authentication/Common/GraphSessionInitializer.cs
```diff
@@ -27,5 +31,34 @@ internal static GraphSession CreateInstance(IDataStore dataStore = null)
                 DataStore = dataStore ?? new DiskDataStore()
             };
         }
+        /// <summary>
+        /// Initializes <see cref="GraphSession"/>. with Output via Cmdlet methods
+        /// </summary>
+        /// <param name="cmdLet"></param>
+        public static void InitializeSession(PSCmdlet cmdLet)
```
peombwa, line 38: We should also remove `public static void InitializeSession()` method in [line 18](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/668/files#diff-b8fd14ba87f272ebb3b25a6c432fddc2666b0f20e8880c2558cd97e44feff1baR18-R21) since it has been replaced this method and is no longer being used.

georgend: Changed it slightly from your recommendation to enable passing a custom writer.

src/Authentication/Authentication.Core/Common/GraphSession.cs
```diff
@@ -200,5 +200,9 @@ public static void Reset()
                 throw new InvalidOperationException(ErrorConstants.Codes.SessionLockWriteDisposed, disposedException);
             }
         }
+        /// <summary>
+        /// Provides Access to output methods provided by the Cmdlet
+        /// </summary>
+        public IPSGraphOutputWriter Output { get; set; }
```
peombwa, line 206:
```suggestion
        public IPSGraphOutputWriter OutputWriter { get; set; }
```

src/Authentication/Authentication/Common/GraphSessionInitializer.cs
```diff
@@ -27,5 +25,43 @@ internal static GraphSession CreateInstance(IDataStore dataStore = null)
...
+                    cmdLet.WriteError(errorRecord);
+                }
+            };
+            InitializeSession(writer);
+        }
+        /// <summary>
+        /// Initializes <see cref="GraphSession"/>. with Output via Cmdlet methods
+        /// </summary>
+        /// <param name="writer"></param>
+        internal static void InitializeSession(IPSGraphOutputWriter writer)
```
peombwa, line 57:
```suggestion
        internal static void InitializeSession(IPSGraphOutputWriter outputWriter)
```

src/Authentication/Authentication.Core/ErrorConstants.cs
```diff
@@ -15,6 +15,7 @@ internal static class Codes
             internal const string SessionLockWriteDisposed = "sessionLockWriteDisposed";
             internal const string SessionLockWriteRecursion = "sessionLockWriteRecursion";
             internal const string InvalidJWT = "invalidJWT";
+            internal const string OutputNotInitialized = nameof(OutputNotInitialized);
```
ddyett, line 18: seems there's a pattern of naming with lower case.  should stay consistent.

georgend: I would propose we take advantage of `nameof` and make the rest upper case?

ddyett: could someone have taken a dependency on the existing strings?  Example, they look for the a specific error message including the string and didn't handle case sensitivity.

peombwa: +1 to what @ddyett said.

@georgend Having error codes in camelCase is also standard even for our [.NET SDK](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dev/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs#L9). We do use camelCasing in PowerShell to align with the casing in .NET Core SDK since we do take a dependency on it and have cases where we pass-through exceptions to PowerShell's error stream.

georgend: @peombwa Thanks for the pointer.

peombwa (approved):
LGTM

PR #709, opened 2021-06-21
Added initial tests for Find-MgGraphPermissions

FehintolaObafemi:
Also added unit test to verify the module's command exports


src/Authentication/Authentication/test/Microsoft.Graph.Authentication.Tests.ps1
```diff
@@ -0,0 +1,39 @@
+Describe "The Microsoft.Graph.Authentication module" {
```
adamedx, line 1: Missing license header.

adamedx (changes requested):
Looks

adamedx (approved):
Thank you for the tests.

PR #722, opened 2021-06-25
 Example Handling Infrastructure and Generated Docs

georgend:
Example Handling Infrastructure and Generated Docs.
Uses a custom version of [autorest.powershell](https://github.com/Azure/autorest.powershell/pull/794).
Changes to Pipeline to pass in Exclude Example Templates flag.


peombwa:
@georgend, consider splitting this into multiple PRs? This PR is too large to load on my machine for a review. We should have a separate PR with docs updates that's separate from infrastructure changes.

PR #735, opened 2021-07-01
Fehintola obafemi/permission expected output

FehintolaObafemi:


src/Authentication/Authentication/custom/common/datasample.json
```diff
@@ -0,0 +1,5424 @@
...
+      "userConsentDisplayName": "Submit application packages to your organization's catalog and cancel pending submissions",
+      "value": "AppCatalog.Submit",
+      "adminConsentDescription": "Allows the app to submit application packages to the catalog and cancel submissions that are pending review on behalf of the signed-in user.",
+      "type": "User",
+      "adminConsentDisplayName": "Submit application packages to the catalog and cancel pending submissions",
+      "isEnabled": true,
+      "id": "3db89e36-7fa6-4012-b281-85f3d9d9fd2e"
+    },
+    {
+      "userConsentDescription": "Allows the app to read the Teams apps that are install
```
adamedx, line 5405: Let's change this to a random guid, i.e. use `new-guid | select -ExpandProperty guid | clip` to add a guid to your clipboard and paste it in. Basically any id's or names that aren't related to permissions should be replaced to avoid exporting private data from an organization.

adamedx, line 5415: Let's do the `new-guid` thing for this also.

adamedx (changes requested):
Minor changes, also need to scrub organizational identifiers from the sample json.

PR #753, opened 2021-07-12
Cleanup Workflows naming scheme

georgend:
Rename workflows appropriately.
- create-helpdocs-pr-workflow.yml -> Creates a PR with Help Docs and Examples generated during build/
- create-openapidocs-pr-workflow.yml -> Creates a PR with Weekly Downloaded OpenAPIDocs


.github/workflows/generate-openapidocs-pr-workflow.yml
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 # This is a basic workflow to help you get started with Actions
 
-name: WeeklyOpenApiDocsDownload
+name: "Generate Weekly OpenAPIDocs Download PR"
```
peombwa, line 5: Let's use "create" instead of "generate" here since are opening a PR for review. The file name should also be renamed to `create-openapidocs-pr-workflow.yml`
```suggestion
name: "Create Weekly OpenAPIDocs Download PR"
```

.github/workflows/generate-helpdocs-pr-workflow.yml
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 # This is a basic workflow to help you get started with Actions
 
-name: "Generate Docs PR"
+name: "Generate Help Docs PR"
```
peombwa, line 5: Let's use "create" instead of "generate" here since are opening a PR for review. The file name should also be renamed to `create-helpdocs-pr-workflow.yml`

```suggestion
name: "Create Help Docs PR"
```

peombwa (approved):

PR #761, opened 2021-07-14
Fehintola obafemi/improved permissions

FehintolaObafemi:
Made some file name changes.


src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -3,6 +3,7 @@
 # ------------------------------------------------------------------------------
 
 $permissionsMsGraphServicePrincipal = $null
+$permissionsPulledFromMgGraphRequest = $null
```
adamedx, line 6: You can omit this until we have code that uses it.

adamedx (approved):

PR #771, opened 2021-07-20
Test files

FehintolaObafemi:
Added some cosmetic changes and included test cases to verify Find-MgGraphPermission's functionality


src/Authentication/Authentication/build-module.ps1
```diff
@@ -92,7 +92,7 @@ Copy-Item -Path "$cmdletsSrc/$ModulePrefix.$ModuleName.psd1" -Destination $outDi
 Copy-Item -Path "$cmdletsSrc/StartupScripts" -Filter *.ps1 -Recurse -Destination $outDir
 
 # Copy custom commands
-Copy-Item -Path "$cmdletsSrc/custom" -Filter *.ps1 -Recurse -Destination $outDir
+Copy-Item -Path "$cmdletsSrc/custom" -Recurse -Destination $outDir
```
adamedx, line 95: I don't think this is 100% safe -- you can easily pick up temp files left by editors for instance. If the goal is to include the test ps1 file, just add another `copy-item` line referencing the directory that hosts the files -- limiting the file type will keep you from picking up unwanted files. That said, I'm not sure what file(s) we're trying to grab here...

FehintolaObafemi: This was to include the json file in the build process

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -2,12 +2,15 @@
...
+
+$permissionsPulledFromMgGraphRequest = $null
 
 function Permissions_GetPermissionsData {
 
+    $permissionsMsGraphServicePrincipal = $null
+
     # 2. Making a REST request to MS Graph
-    if ($null -eq $script:permissionsMsGraphServicePrincipal){
+    if ($null -eq $permissionsMsGraphServicePrincipal){
```
adamedx, line 13: Now is probably a good time to say that the `permissionsMSGraphServicePrincipal` variable should match the naming convention of the functions, i.e. can you rename this variable throughout to use an underscore as in `permissions_MsGraphServicePrincipal`? That gets across the fact that this variable is considered a (static) 'member' of the `Permissions` 'class'.

src/Authentication/Authentication/build-module.ps1
FehintolaObafemi: I noticed an error in the build while I tried to add a new Copy-Item line which is why I opted to omitting it

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -2,15 +2,170 @@
...
+            Mock Invoke-MgGraphRequest{
+                $permissionData
+            }
+
+            { Find-MgGraphPermission mail | Out-Null } | Should -Not -Throw
+            Assert-MockCalled Invoke-MgGraphRequest
         }
 
+        It "Returns nothing and throws no exception if a search string is specified and there is no match" {
+            { Find-MgGraphPermission 'Microsoft is the greatest company' | Out-Null } | Should -Not -Throw
```
adamedx, line 112: Heh! I like this, but we do need to avoid people feeling we aren't neutral here since this is an OSS community project maintained by Microsoft but "owned" in a sense by the larger community. That said, it might be an opportunity to weigh in on where the best jollof rice comes from...

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -2,15 +2,170 @@
...
+        }
     }
 
-}
\ No newline at end of file
+    Context "When executing the command when initially not having access to read the service prinicpal but later having access" {
+
+        BeforeEach {
+            Mock Invoke-MgGraphRequest {
+
```
adamedx, line 128: For consistency you can remove the extra spaces here -- the indentions I think provide enough separation for readability.

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -2,15 +2,170 @@
...
+        It "Retrieves the expected set of delegated and app-only permissions when a search string is specified" {
+            { Find-MgGraphPermission mail | Out-Null } | Should -Not -Throw
+            Assert-MockCalled Invoke-MgGraphRequest
+
+            Mock Invoke-MgGraphRequest{
+                $permissionData
+            }
+
+            { Find-MgGraphPermission mail | Out-Null } | Should -Not -Throw
```
adamedx, line 107: Now you need to do more than just check the data. You must also check the results from the test file. Currently you're not checking for instance how many permissions get returned when you searched for `mail`. You'll need to check that the results match the expectations.

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -2,15 +2,170 @@
...
+        }
+
+        It "Returns nothing and throws no exception if a search string is specified and there is no match" {
+            { Find-MgGraphPermission 'Microsoft is the greatest company' | Out-Null } | Should -Not -Throw
+            Assert-MockCalled Invoke-MgGraphRequest
+        }
+    }
+
+    Context "When executing the command after initially not having a connection to MS Graph but getting it later" {
+
```
adamedx, line 77: One more thing -- in these tests you'll need some sort of state that tells you whether you are using the mocked service principal data or the static file from the source code. An easy way to check would be for your mock of `Invoke-MgGraphRequest` to also set some external state (e.g. a script variable) that you reset before you invoke `Find-MgGraphRequest`. If that mock changes it, you know for sure your second mock got called for instance. We may need to add deeper logic to test state variables of the code itself to see that they are also set correctly.

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -2,15 +2,170 @@
...
+                $permissionData
+            }
+
+            { Find-MgGraphPermission | Out-Null } | Should -Not -Throw
+            Assert-MockCalled Invoke-MgGraphRequest
+
+        }
+
+        It "Retrieves the expected set of delegated and app-only permissions when a search string is specified" {
+            { Find-MgGraphPermission mail | Out-Null } | Should -Not -Throw
```
adamedx, line 100: Rather than check to see if it doesn't throw, just do `$permissions  Find-MgGraphPermission mail`. Then your next line can be `$permissions.length | Should -Be 4` (or whatever is the number of permissions in the file). And add more lnes like `$permissions.ConsentType | Should -Be Admin` or whatever the right value is. You'll want to pick useful cases and validate that the code is behaving as you expect.

src/Authentication/Authentication/build-module.ps1
adamedx: Can we drill in to what the error was? We should fix that as the approach here could include unwanted files.

src/Authentication/Authentication/custom/common/Permissions.ps1
adamedx: It's still missing the `_` underscore character -- can you make that change?

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -15,12 +15,18 @@ Find permissions for authorization against Microsoft Graph
 
 # Import-Module -Force .\Microsoft.Graph.Authentication.psd1
 
-function Find-MgGraphPermission($search) {
+function Find-MgGraphPermission([string]$search) {
```
adamedx, line 18: Need to add [[cmdletbinding]](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_cmdletbindingattribute?view=powershell-7.1)  and use the `param` syntax for parameters, i.e.

```powershell
function Find-MgGrapPermission {
    [cmdletbinding()]
    param (
       [string] $search
    )
    # the rest of the code for the function.
}

File: src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1 [hunk 9]
```diff
@@ -15,12 +15,18 @@ Find permissions for authorization against Microsoft Graph
 
 # Import-Module -Force .\Microsoft.Graph.Authentication.psd1
 
-function Find-MgGraphPermission($search) {
```
adamedx, line 18: We should use a more standard name like `$SearchString` -- also need to capitalize parameter names.

adamedx (approved):
Thanks for addressing the earlier feedback.

PR #784, opened 2021-07-26
Find-MgGraphPermission: Add formatting

FehintolaObafemi:
I have a concern regarding the final output. Since we are already grouping by the permission types would it not be redundant to then display the permission types as one of the four output columns
Capture


FehintolaObafemi:
Ensured the ps1xml file path was being properly called into the function for formatting purposes.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -22,15 +26,20 @@ function Permissions_GetPermissionsData {
...
 
-            # Write-Host "Getting data from local file"
+            Write-Host "You are not connected to MgGraph, reverting you back to the static file"
             Get-Content $PSScriptRoot/MSGraphServicePrincipalPermissions.json | Out-String | ConvertFrom-Json
         
         } catch [System.Net.Http.HttpRequestException] {
 
-            # Write-Host "Getting data from local file"
+            Write-Host "Inadequate access required to view MgGraph service principals, reverting you back to the static file"
             Get-Content $PSScriptRoot/MSGraphServicePrincipalPermissions.json | Out-String | ConvertFrom-Json
```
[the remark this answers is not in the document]
FehintolaObafemi: There are two different test cases the first being not signing into mgGraph and the second being not having adequate access. The use of different catch blocks for this is to properly run tests and verify it's functionality

> I don't think it makes sense to treat these catch blocks independently. Probably should use just one catch block and not duplicate code.
> 
> catch multiple types like this:
> `catch [System.IO.DirectoryNotFoundException],[System.IO.FileNotFoundException]`

adamedx:
Can you show an example of the output in the PR description? This will help for posterity when people are looking at previously merged changes and want to understand the purpose / goal.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,47 @@
...
             
             if ($null -ne $result) {
                 $result | select-object -expandproperty value 
+                $fromInvokeMgGraphRequest = $true
             }
 
         } catch [System.Management.Automation.ValidationMetadataException] {
 
-            # Write-Host "Getting data from local file"
+            $requestError = "Connection to MgGraph required. This can be done by using 'Connect-MgGraph'"
```
adamedx, line 30: Is this necessary? We could just record the error rather than using a custom string.. Also, it's not clear if we need to support localization. If we're going to add custom messages, we will need to plug in to localization framework. Might be something that can be done in a future PR.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,47 @@
...
             Get-Content $PSScriptRoot/MSGraphServicePrincipalPermissions.json | Out-String | ConvertFrom-Json
         
         } catch [System.Net.Http.HttpRequestException] {
 
-            # Write-Host "Getting data from local file"
+            $requestError = "Administator access to the tenent required."
```
adamedx, line 35: typo `Administator` -> `Administrator`. But I still think we should avoid custom error messages unless we can localize.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -40,16 +51,28 @@ function Permissions_GetPermissionsData {
     # make sure the parsed permissions are exported properly
     @{
         oauth2 = $msOauth;
-        appRoles = $msAppRoles
+        appRoles = $msAppRoles;
```
adamedx, line 54: The semicolon is unnecessary and should be removed.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -64,33 +87,57 @@ function Permissions_GetOauthData {
...
+    ForEach ($approle in $msAppRoles) {
 
-        $entry = [ordered] @{
-            "Id" = $oauth2grant.id
-            "PermissionType" = $oauth2grant.origin
-            "Name" = $oauth2grant.value
-            "Description" = $oauth2grant.description
+        $consent = If ($approle.origin -eq "Application") { 
+        
+            "Admin"
```
adamedx, line 124: Let's use "Administrator," not "Admin. Also, no need for empty lines before and after.

adamedx:
@FehintolaObafemi , I don't see the unit tests in this PR -- did you push these?

src/Authentication/Authentication/custom/common/Permissions.ps1
FehintolaObafemi: I have gotten rid of the empty lines. I left it as admin so that it can match the consent values for the oauth2

adamedx: I agree with your take on `admin` -- our documentation actually uses the term `Admin` for this exact instance of data, and as you point out that is also consistent with the REST API for `oauth2PermissionGrant`. Also, since this is PowerShell, it straddles the world of Presentation UX + Automation / API, so it's ok for it to make a choice that's consistent with an API but not the same as a more "official" presentation that would be found in Azure Portal, which does not support automation.

[Our docs do indeed use admin instead of administrator](https://docs.microsoft.com/en-us/graph/permissions-reference#access-reviews-permissions)

FehintolaObafemi:
I have included test cases for the online switch instances to confirm that it works as expected. I also updated the static file to show the appropiate id's for each command.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,49 @@
 # ------------------------------------------------------------------------------
 #  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 # ------------------------------------------------------------------------------
-
-
-$permissionsPulledFromMgGraphRequest = $null
+Update-FormatData -PrependPath $PSScriptRoot\..\..\Microsoft.Graph.Authentication.format.ps1xml
```
adamedx, line 4: This should not be done here -- let's do this in the .psd1 file -- just add this path to `formatsToProcess`.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,49 @@
...
+
+    if (($null -eq $permissions_MsGraphServicePrincipal) -or ($null -ne $permissions_MsGraphServicePrincipal -and $fromInvokeMgGraphRequest -eq $false)) {
         $script:permissions_MsGraphServicePrincipal = try {
 
             # Write-Host "Getting data from web service"
             $result = Invoke-MgGraphRequest -method GET 'https://graph.microsoft.com/v1.0/servicePrincipals?filter=appId eq ''00000003-0000-0000-c000-000000000000''' 
             
             if ($null -ne $result) {
                 $result | select-object -expandproperty value 
+                $script:fromInvokeMgGraphRequest = $true
```
adamedx, line 25: This variable should be initialized at script scope to `$null` -- just do that at the top of the file.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,49 @@
...
 
     $permissions_MsGraphServicePrincipal = $null
+    $requestError = $null
+    $fromInvokeMgGraphRequest = $false
 
     # 2. Making a REST request to MS Graph
-    if ($null -eq $permissions_MsGraphServicePrincipal){
+
+    if (($null -eq $permissions_MsGraphServicePrincipal) -or ($null -ne $permissions_MsGraphServicePrincipal -and $fromInvokeMgGraphRequest -eq $false)) {
         $script:permissions_MsGraphServicePrincipal = try {
```
adamedx, line 18: This should be initialized to `$null` at the top of the file.

src/Authentication/Authentication/custom/common/Permissions.ps1
adamedx: Also, please add `Set-StrictMode -Version 6.0` to the top of the every `.ps1` file that we created. That should help catch issues with variables that are used prior to initialization.

src/Authentication/Authentication/custom/common/Permissions.ps1
FehintolaObafemi: They had been initialized at the top of the file

src/Authentication/Authentication/custom/common/Permissions.ps1
FehintolaObafemi: They had been initialized at the top of the file

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -18,18 +18,23 @@ Find permissions for authorization against Microsoft Graph
 function Find-MgGraphPermission {
     [cmdletbinding()]
     param (
-        [string] $SearchString
```
adamedx, line 21: The comment with the synopsis field should be immediately on top of the `Find-MgGraphPermission` function.

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -43,30 +48,38 @@ Describe "the Find-MgGraphPermission Command" {
...
-
             }
         }
 
         It 'Executes successfully with no parameters' {
             { Find-MgGraphPermission | Out-Null } | Should -Not -Throw
             Assert-MockCalled Invoke-MgGraphRequest
+            $permissions_MsGraphServicePrincipal | Should -Not -Be $null
+            $fromInvokeMgGraphRequest | Should -Be $false
+            { Find-MgGraphPermission -Online | Out-Null } | Should -Throw 'mock connection error message'
```
adamedx, line 82: The `-online` should be a dedicated test case.

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -75,40 +88,48 @@ Describe "the Find-MgGraphPermission Command" {
...
 
             Mock Invoke-MgGraphRequest{
```
adamedx, line 124: You might want to add a comment here that this is how you are "restoring" the connection that was set to fail in the `BeforeAll`

src/Authentication/Authentication/custom/common/Permissions.ps1
adamedx, line 25: Also, thi should be named with the `permissions_` prefix to show that this variable is part of the `Permissions` class.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,49 @@
...
+        [bool] $online
+    )
 
     $permissions_MsGraphServicePrincipal = $null
+    $requestError = $null
+    $fromInvokeMgGraphRequest = $false
 
     # 2. Making a REST request to MS Graph
-    if ($null -eq $permissions_MsGraphServicePrincipal){
+
```
adamedx, line 16: Would be great to preface this with a comment explaining the logic with human language and the reasons for it.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -1,36 +1,49 @@
...
+    if (($null -eq $permissions_MsGraphServicePrincipal) -or ($null -ne $permissions_MsGraphServicePrincipal -and $fromInvokeMgGraphRequest -eq $false)) {
```
adamedx, line 17: Need to check `$online` and do this if `$online` is specified in all cases -- be careful though not to overwrite existing sp data until after you're successful!

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -24,6 +24,8 @@ Describe "the Find-MgGraphPermission Command" {
         It 'Executes successfully with no parameters' {
             { Find-MgGraphPermission | Out-Null } | Should -Not -Throw
             Assert-MockCalled Invoke-MgGraphRequest
```
adamedx, line 26: Should you assert that it was called exactly 1 time with `-Times 1`?

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -128,12 +152,18 @@ Describe "the Find-MgGraphPermission Command" {
             $test = Find-MgGraphPermission 'ReadWrite'
             $test | Should -Not -Be $null
             $test.length | Should -Be 4
+            $permissions_MsGraphServicePrincipal | Should -Not -Be $null
+            $fromInvokeMgGraphRequest | Should -Be $true
+            { Find-MgGraphPermission 'Readwrite' -Online | Out-Null } | Should -Not -Throw
```
adamedx, line 157: Should have an `Assert-MockCalled` since you defined one here.

Issue #704, opened 2021-06-17
Proposal: Find-MgGraphPermission command to enable use of application API


adamedx:
Find-MgGraphPermission Command
==============================

## Problem

The Microsoft Graph PowerShell SDK application requires users to have domain knowledge of both the semantics and syntax of Microsoft Graph API permissions used to authorize access to the API. There are two key permissions scenarios that present a challenge for users:

1. How do I find the values to supply to the permission-related parameters of commands like `New-MgApplication` and other application and consent related commands? Commands that allow the user to specify permissions like `User.Read` in order to grant those permissions to an application or ensure that the application requests those permissions offer a parameter to identify the permission(s) in question. While the user will be familiar with those permissions through their documented names like `User.Read` that are also used in many other user-facing contexts, the commands require that they be referred to not through the human readable name but by using a GUID identifier. Unfortunately, the identifiers are not documented, so there is no way for a user to tranlsate the intention of say granting an application `User.Read` to parameters that can be supplied to an MS Graph PowerShell SDK command-line. The workaround is for the user to perform the task using the Azure Portal which *does* allow the user to specify the human-readabl name, and once the application is configured other commands that read the configured permissions could return the GUIDs, allowing the user to "reverse-engineer" the friendly-name to GUID mappings.
2. What permissions should I ask for when I invoke the mandatory `Connect-MgGraph` command? To use Microsoft Graph PowerShell SDK to access Microsoft Graph, users must sign in to an Azure Active Directory application using the `Connect-MgGraph` command. Users must explicitly specify the permissions required for the API they are accessing when issuing the command. If a user fails to include required permissions when issuing the `Connect-MgGraph` command and those permissions have not already been consented to the application / user, subsequent commands that access the APIs authorized by those permissions will fail.
3. Currently PowerShell commands and scripts, including those implemented with Microsoft Graph PowerShell SDK itself have no way of validating user input that refers to permissions or providing "auto-complete" user experiences to help users accurately supply input to commands

Note that these cases are listed in priority order -- the first in particular is considered a blocking issue. The other two still have significant impact on user experience, but there are some less-productive workarounds that users may employ.

## Solution

All three issues can be resolved through a solution that addresses the first issue, which is itself a blocking issue

### Find-MgGraphPermission

```powershell
function Find-MgGraphPermission {
    [cmdletbinding(positionalbinding=$false)]
    param(
        [parameter(position=0, parametersetname='searchspec')]
[...]

PR #809, opened 2021-08-06
Find-MgGraphPermission

FehintolaObafemi:
This change adds the `Find-MgGraphPermission` command to the `Microsoft.Graph.Authentication` module as part of #704. See that issue for details on the functionality and use cases. Here are the implementation highlights:

* The `Find-MgGraphPermission` command is implemented as a custom script command using PowerShell itself, not C#
* A custom type `Microsoft.Graph.Custom.Permission` is defined as the output type of `Find-MgGraphPermission` so that other commands like `Select-Object` can implement auto-complete for the output of `Find-MgGraphPermission`
* The change also includes new output formatting in the `ps1xml` definition for `Microsoft.Graph.Custom.Permission`
* Existing sample scripts for `Application` and `Connect` have been updated to include relevant use cases involving `Find-MgGraphPermission`
* Documentation is included with the command as part of comment-based documentation
* Unit test coverage is accomplished through a set of *Pester* test cases

To understand the expected behavior, you can review:
* [The original Find-MgGraphPermission proposal / specification](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/FehintolaObafemi/permission/src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1)
* [The command documentation](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_methods?view=powershell-7.1#begin)
* [The unit tests](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/FehintolaObafemi/permission/src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1)

As noted in the original proposal, the permissions data is sourced from Microsoft Graph itself. However, many users will not have the privileges required to read it (you must be able to read the Microsoft Graph service principal object in your Azure Active Directory organization; company admins will have no problem with this, but many lower-privileged users won't be able to. Due to this, there is a fallback to snapshot of that service principal included in the code. For 99% of cases, the fallback will be fine as the newest permissions are unlikely to be those that someone is looking for. We do need to periodically update the file though.

Note that the command also includes additional documented capabilities that were not part of the original proposal -- these were added to simplify automation cases where you where partial matches and non-determinism are unacceptable. The additional filtering allows the set of results to be narrowed down to just one result (or none!) which is required for scenarios such as translating a known friendly permission name to the unique id for use with Microsoft Graph REST APIs or commands for those APIs.


adamedx:
@peombwa , I can't find a log of the Pester test runs in the repository -- do you have a pointer to this?

peombwa:
> I can't find a log of the Pester test runs in the repository -- do you have a pointer to this?

@adamedx, the Pester test run results can be found [here](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=53924&view=ms.vss-test-web.build-test-results-tab). @georgend, would you know why Invoke-MgGraphRequest tests a failing here?

cc\ @FehintolaObafemi.

adamedx:
> > I can't find a log of the Pester test runs in the repository -- do you have a pointer to this?
> 
> @adamedx, the Pester test run results can be found [here](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=53924&view=ms.vss-test-web.build-test-results-tab). @georgend, would you know why Invoke-MgGraphRequest tests a failing here?
> 
> @FehintolaObafemi, you can ignore the failing `Invoke-MgGraphRequest` tests since they are not related to your changes.

Thanks @peombwa -- it seems I don't have permission to see the test results, even though I'm signed in to Azure DevOps. Any way to fix that? If not, can you validate that the tests we added are indeed running? The unit test file is linked in the PR description.

samples/9-Applications.ps1
```diff
@@ -26,6 +26,7 @@ $app3 = New-MgApplication -displayName "ImplicitWebApp" `
         }  
 
 # Create an registration for an ASP.NET Web App
+$scopeId_UserRead = Find-MgGraphPermission User.Read -ExactMatch -PermissionType Delegated | Select-Object -ExpandProperty Id
```
adamedx, line 29: @peombwa you'll want to sign off on having the sample changed to use the new command.

samples/9-Applications.ps1
```diff
@@ -76,18 +77,36 @@ $Certificate = Get-ChildItem -Path "Cert:\CurrentUser\My\$CertificateThumbprint"
 # Graph resource Id
 $GraphResourceId = "00000003-0000-0000-c000-000000000000"
 
-# Graph permissions constants
-$UserReadAll = @{ Id = "df021288-bdef-4463-88db-98f22de89214"; Type = "Role" }
-$GroupReadAll = @{ Id = "5b567255-7703-4780-807c-7be8301ae99b"; Type = "Role" }
-$MailboxSettingsRead = @{ Id = "40f97065-369a-49f4-947c-6a255697ae91"; Type = "Role" }
-$MailSend = @{ Id = "b633e1c5-b582-4048-a93e-9f11b44c7e96"; Type = "Role" }
+# Show friendly Graph permission names given their unique identifiers
```
adamedx, line 80: Another big change to the sample @peombwa

samples/2-ConnectToGraph.ps1
```diff
@@ -6,6 +6,9 @@ Connect-Graph
 # Try to Get-User
 Get-MgUser
 
+# Search for delegated permissions related to sites
```
adamedx, line 9: @peombwa goal here is to show you can use the new command to find permissions, and also to get to the permissions reference help.

samples/2-ConnectToGraph.ps1
```diff
@@ -17,3 +20,7 @@ Connect-Graph -Scopes "User.Read","User.ReadWrite.All","Mail.ReadWrite",`
 
 # Forget all access tokens
 Disconnect-Graph
+
+# Launch detailed permissions documentation
+Get-Help Find-MgGraphPermission -Online
```
adamedx, line 25: @msewaweru , this works because the .LINKS field in the comment help has exactly one entry, and it's the URI to the permissions docs. That means we can't provide references to other commands in .LINKS though -- hope that's ok.

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -0,0 +1,226 @@
...
+        MemberType = 'NoteProperty'
+        MemberName = $_
+        Value = $null
+    }
+
+    Update-typedata @typeParameters -force
+}
+
+<#
+.SYNOPSIS
```
adamedx, line 23: @msewaweru this is the documentation section -- I believe this will be used to generate the online docs...

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -0,0 +1,226 @@
...
+Name        : MailboxSettings.ReadWrite
+Description : Allows the app to read, update, create, and delete your mailbox settings
+
+This example outputs the set of delegated permissions that match the 'mailbox' SearchString parameter. The Where-Object
+command is used to filter the permissions to only the delegated permissions through the PermissionType property. The filtered
+result is piped to the Format-List command so that the output of the Description field is not truncated as it would be
+in the default table view.
+
+.LINK
+https://docs.microsoft.com/en-us/graph/permissions-reference
```
adamedx, line 168: This allows `Get-Help -Online Find-MgGraphPermission` to actually point at the permissions reference rather than help for `Find-MgGraphPermission`. If that's not ok, we can change this.

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -0,0 +1,226 @@
...
+that your application or script follows the best security pratice of using the least privilege necessary
+are found in the Microsoft Graph Permissions Reference documenation at https://docs.microsoft.com/en-us/graph/permissions-reference.
+
+In order to search for the specified permissions, Find-MgGraphPermission uses Microsoft Graph itself
+to obtain the latest complete set of permissions; if it is not able to access Microsoft Graph, the command
+then uses its most recent copy of the information as the source of data.
+
+Note that the results of Find-MgGraphPermission grouped by Delegated and Application permission types, and then sorted
+by permission name. They are not sorted by least-privileged access, so when multiple permissions appear to satisfy your
+script or application's requirements, consult the Microsoft Graph Permissions Reference documenation to identify the
```
TannerTrombley, line 51:
```suggestion
script or application's requirements, consult the Microsoft Graph Permissions Reference documentation to identify the
```

src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1
```diff
@@ -0,0 +1,226 @@
...
+least-privileged permission for your use case.
+
+.PARAMETER SearchString
+The SearchString parameter allows you to specify a string such as 'user' or 'mail' that represents the subject or domain
+of the permission you're searching for. Since permissions usually have names such as 'User.Read' or 'Mail.ReadWrite', the
+command uses the SearchString parameter to return all permissions that contain the value specified for SearchString in the
+name of the permission.
```
TannerTrombley, line 58: @FehintolaObafemi Clarification: I thought you said this searches the description field as well. I don't think that is the case, but just want to make sure.

FehintolaObafemi: @TannerTrombley Adam and I opted to excluding the description field search.

peombwa:
> Thanks @peombwa -- it seems I don't have permission to see the test results, even though I'm signed in to Azure DevOps. Any way to fix that? If not, can you validate that the tests we added are indeed running? The unit test file is linked in the PR description.

AzDO is picking up the Find-MgGraphPermission Pester tests and running them as expected. However, the following test is failing with:
``` powershell
The Find-MgGraphPermission Command.When executing the command using a constrained set of permissions returned by MS Graph and there is a connection.Should return null and not throw an exception if ExactMatch is specified and there is no match

Expected no exception to be thrown, but an exception "No results were found that exactly matched the specified permission 'IDontExist'" was thrown from C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119 char:15
+ …           { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out- …
+               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
at { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out-null } | Should -Not -Throw, C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119
```

adamedx:
> > Thanks @peombwa -- it seems I don't have permission to see the test results, even though I'm signed in to Azure DevOps. Any way to fix that? If not, can you validate that the tests we added are indeed running? The unit test file is linked in the PR description.
> 
> AzDO is picking up the Find-MgGraphPermission Pester tests and running them as expected. However, the following test is failing with:
> 
> ```powershell
> The Find-MgGraphPermission Command.When executing the command using a constrained set of permissions returned by MS Graph and there is a connection.Should return null and not throw an exception if ExactMatch is specified and there is no match
> 
> Expected no exception to be thrown, but an exception "No results were found that exactly matched the specified permission 'IDontExist'" was thrown from C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119 char:15
> + …           { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out- …
> +               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
> at { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out-null } | Should -Not -Throw, C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119
> ```
> 
> cc\ @FehintolaObafemi

Interesting -- then my question why are the checks passing? Shouldn't this failure cause the checks to all fail and turn red?

I'm not sure why this would fail unless the `$ErrorActionPreference` is set to something non-standard. I have an idea how to compensate for that...

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
```diff
@@ -0,0 +1,417 @@
...
+            $delegatedIndex = 0
+
+            $delegatedPermissions | ForEach-Object {
+                $delegatedMatches[$delegatedIndex++].Name | Should -Be $_
+            }
+        }
+
+        It "Should return null and not throw an exception if ExactMatch is specified and there is no match" {
+            { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out-null } | Should -Not -Throw
+            $ErrorActionPreference = 'continue'
```
peombwa, line 120: `$ErrorActionPreference = 'continue'` should be specified before making the call. This is the cause of https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/809#issuecomment-895325576.
```suggestion
            $ErrorActionPreference = 'continue'
            { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out-null } | Should -Not -Throw
```

samples/2-ConnectToGraph.ps1
```diff
@@ -6,6 +6,9 @@ Connect-Graph
 # Try to Get-User
 Get-MgUser
 
+# Search for delegated permissions related to sites
+Find-MgGraphPermission sites
```
peombwa, line 10: The sample below returns both delegated and application permissions. PermissionType should be specified.

```suggestion
# Search for delegated permissions related to sites
Find-MgGraphPermission sites -PermissionType Delegated
```

samples/9-Applications.ps1
peombwa: This looks good! This is a good example of how `Find-MgGraphPermission` can be used to aid in scope discovery.

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -0,0 +1,102 @@
...
+            if ( $restResult ) {
+                $_permissions.msGraphServicePrincipal = $restResult | Select-Object -ExpandProperty value
+                $_permissions.isFromInvokeMgGraphRequest = $true
+            }
+        } catch [System.Management.Automation.ValidationMetadataException], [System.Net.Http.HttpRequestException] {
+            # We can't get the data from MS Graph, so just use a local static (possibly stale) copy
+            $_permissions.msGraphServicePrincipal = Get-Content $PSScriptRoot/MSGraphServicePrincipalPermissions.json | Out-String | ConvertFrom-Json
+            if ( $online ) {
+                throw
+            }
```
peombwa, line 48: To be memory efficient, you should throw the exception first (when `-Online` is true) before reading the file.
```suggestion
            if ( $online ) {
                throw
            }
            # We can't get the data from MS Graph, so just use a local static (possibly stale) copy
            $_permissions.msGraphServicePrincipal = Get-Content $PSScriptRoot/MSGraphServicePrincipalPermissions.json | Out-String | ConvertFrom-Json
```

src/Authentication/Authentication/test/Find-MgGraphPermission.Tests.ps1
adamedx: > `$ErrorActionPreference = 'continue'` should be specified before making the call. This is the cause of [#809 (comment)](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/809#issuecomment-895325576).

Yeah, we ended up doing that @peombwa as a workaround. But technically this is incorrect -- the default value for `$ErrorActionPreference` is `Continue` as stated in the [documentation](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#erroractionpreference). As far as validating the functionality our workaround will get the job done (maybe even passing that erroraction preference as a parameter instead of using the variable), but it seems like something in the test infrastructure is altering the standard powershell preferences.

samples/2-ConnectToGraph.ps1
georgend: @maisarissi given the pointer by @adamedx it means we cannot use related links as we had intended to store Survey Links.

adamedx: You may want to sync with someone from the https://github.com/powershell/powershell or other experts on dcs + PowerShell project to confirm the behavior I observed.

peombwa:
> > Thanks @peombwa -- it seems I don't have permission to see the test results, even though I'm signed in to Azure DevOps. Any way to fix that? If not, can you validate that the tests we added are indeed running? The unit test file is linked in the PR description.
> 
> AzDO is picking up the Find-MgGraphPermission Pester tests and running them as expected. However, the following test is failing with:
> 
> ```powershell
> The Find-MgGraphPermission Command.When executing the command using a constrained set of permissions returned by MS Graph and there is a connection.Should return null and not throw an exception if ExactMatch is specified and there is no match
> 
> Expected no exception to be thrown, but an exception "No results were found that exactly matched the specified permission 'IDontExist'" was thrown from C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119 char:15
> + …           { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out- …
> +               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
> at { Find-MgGraphPermission -ExactMatch IDontExist 2>&1 | out-null } | Should -Not -Throw, C:\a\1\s\src\Authentication\Authentication\test\Find-MgGraphPermission.Tests.ps1:119
> ```

@adamedx, I've figured out why the failing tests were not causing the checks to fail. You can find the PR to fix this [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/819).

src/Authentication/Authentication/StartupScripts/ExportCustomCommands.ps1
```diff
@@ -0,0 +1,22 @@
...
+    }
+}
+
+# Export custom script commands without removing the
+# binary cmdlets. Custom script commands are functions,
+# the cmdlets are.. cmdlets. We must explicitly specify
+# both functions and cmdlets at export; if only one of
+# these classes is specified, nothing of the other
+# class will be exported.
+Export-ModuleMember -Function * -Cmdlet *
```
peombwa, line 21: Just an FYI Exporting module members like this will suppress currently exported cmdlet aliases such as `Connect-Graph`. The Pester tests to confirm this can be found [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/fc50dfe002d11adfbe70fae1d8d6f12a8d8a24e5/src/Authentication/Authentication/test/Microsoft.Graph.Authentication.Tests.ps1#L58-L68). I've fixed this in my PR by adding [`Get-ModuleCmdlet`](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/816/files#diff-1850068b7ea78c6849651f57381ce2283f70405350c22cd4f6ad75d7a632738fR16-R18.) and [`Get-ScriptCmdlet`](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/816/files#diff-387044d693da97c2f19f907463d42ebc596aee909532d409ebc1615ff59213ccR11) helper cmdlets to aid in exporting functions, cmdlets, and aliases.

This will be fixed https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/816 when we merge the 2 PRs.

peombwa (dismissed):
Other than the issue highlighted [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/809/files#r687265838) (which has been fixed by PR https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/816), everything else looks good to me!

src/Authentication/Authentication/StartupScripts/ExportCustomCommands.ps1
adamedx: > Other than the issue highlighted [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/809/files#r687265838) (which has been fixed by PR #816), everything else looks good to me!

Awesome, thank you @peombwa! What are the next steps? It sounds like this is in your queue to merge and you'll be the person to handle merging this (along with the related PR's)?

peombwa: @adamedx, That's right. I'll merge the PR later today (to give others time to review it) then schedule everything for a 1.7.0 release somewhere next week.

@FehintolaObafemi, thank you for the excellent contribution!! I'm quite sure our customers will find this command valuable for their scripts.

peombwa (approved):

Issue #274, opened 2020-06-24
Set the property error category when an item is not found, and clean up Activity property


KirkMunro:
If I attempt to retrieve a resource that does not exist, I get an exception. That exception currently shows the category for the error as `InvalidOperation`, which is incorrect. The operation was valid, but the resource was not found.

For example, run this:

```powershell
Get-MgUser -UserId (New-Guid).Guid -ErrorVariable lookupError
$lookupError.CategoryInfo
```

This script results in an error from the first command (you'll get an error because there is no user with that random GUID that you just created), and then it shows the category information for the error. The output from the second command will look something like this:

```none
Category   : InvalidOperation
Activity   : Get-MgUser_Get
Reason     : Exception
TargetName : { UserId = 4854a621-9117-4cd3-999b-5f7ca6882432, Property = , ExpandProperty =  }
TargetType : <>f__AnonymousType0`3
```

The categorization of this error is being done incorrectly, and that makes it difficult for scripters to properly handle a scenario when a resource is not found. For example, you delete a resource and then want to make sure that it was deleted. The category of `InvalidOperation` should be reserved for operations that failed due to the operation being invalid. When a resource is not found, you should use `[System.Management.Automation.ErrorCategory]::ObjectNotFound` as the category instead.

Further, for the `Activity` property, why does it show an extra "_Get" at the end? I invoked `Get-MgUser`. I expect the activity would be simply `Get-MgUser`. What is the significance of the extra verb appended on the end of that string?

peombwa:
Feature is being tracked externally at https://github.com/Azure/autorest.powershell/issues/888.

PR #973, opened 2021-12-10
Properly handle async calls when logging request/response bodies

peombwa:
This PR ensures request/response bodies are always logged when `-Debug` is set.

### DIFF
--- a/src/readme.graph.md
+++ b/src/readme.graph.md
@@ -534,6 +534,11 @@ directive:
         let duplicateDebugRegex = /^(\s*)(WriteDebug\(\$"{id}:.*)/gmi
         $ = $.replace(duplicateDebugRegex, "");
 
+        // catch all exceptions in ProcessRecordAsync.
+        let processAsyncFinallyRegex = /(finally\s*{\s*await \(\(Microsoft\.Graph\.PowerShell\.Runtime\.IEventListener\)this\)\.Signal\(Microsoft\.Graph\.PowerShell\.Runtime\.Events\.CmdletProcessRecordAsyncEnd\);)/gmi
+        let catchAllExceptionImplementation = '((Runtime.IEventListener)this).Signal(Runtime.Events.CmdletException, $"{ex.GetType().Name} - {ex.Message} : {ex.StackTrace}").Wait(); if (((Runtime.IEventListener)this).Token.IsCancellationRequested) { return; } WriteError(new global::System.Management.Automation.ErrorRecord(ex, string.Empty, global::System.Management.Automation.ErrorCategory.NotSpecified, null));'
+        $ = $.replace(processAsyncFinallyRegex, `catch (System.Exception ex){${catchAllExceptionImplementation}}\n$1`);
+
         return $;
       }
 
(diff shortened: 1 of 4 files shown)


tools/Custom/HttpMessageLogFormatter.cs
```diff
@@ -44,7 +44,7 @@ public static string GetHttpResponseLog(HttpResponseMessage response)
             string body = string.Empty;
             try
             {
-                body = (response.Content == null) ? string.Empty : FormatString(response.Content.ReadAsStringAsync().Result);
+                body = (response.Content == null) ? string.Empty : FormatString(response.Content.ReadAsStringAsync().GetAwaiter().GetResult());
             }
             catch { }
```
georgend, line 49: What happens when there is an exception, such as a network error?

peombwa: Transport exceptions will be handled by `OnCmdletException` - https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/bugfixes/SerializationDepth/tools/Custom/Module.cs#L68-L70.
image

Here, we are just logging an empty body if we can't read the body content for whatever reason. The code should still flow, but the logs will have an empty body.

peombwa: I just noticed that AutoREST is not catching HTTP request timeouts. My last commit fixes this by adding a `catch(System.Exception ex)` in `ProcessRecordAsync` method. Only `PipelineStoppedException`, `AggregateException`, `UndeclaredResponseException`, and `RestException` were caught.

src/readme.graph.md
```diff
@@ -534,6 +534,11 @@ directive:
         let duplicateDebugRegex = /^(\s*)(WriteDebug\(\$"{id}:.*)/gmi
         $ = $.replace(duplicateDebugRegex, "");
 
+        // catch all exceptions in ProcessRecordAsync.
+        let processAsyncFinallyRegex = /(finally\s*{\s*await \(\(Microsoft\.Graph\.PowerShell\.Runtime\.IEventListener\)this\)\.Signal\(Microsoft\.Graph\.PowerShell\.Runtime\.Events\.CmdletProcessRecordAsyncEnd\);)/gmi
+        let catchAllExceptionImplementation = '((Runtime.IEventListener)this).Signal(Runtime.Events.CmdletException, $"{ex.GetType().Name} - {ex.Message} : {ex.StackTrace}").Wait(); if (((Runtime.IEventListener)this).Token.IsCancellationRequested) { return; } WriteError(new global::System.Management.Automation.ErrorRecord(ex, string.Empty, global::System.Management.Automation.ErrorCategory.NotSpecified, null));'
```
georgend, line 539: Is there a more specific error category than `ErrorCategory.NotSpecified`?

And for the Signal, a more specific exception than `Runtime.Events.CmdletException`?

georgend (approved):

src/readme.graph.md
peombwa: `NotSpecified` is best suited for this scenario since this is a catch-all uncaught exception. AutoREST already has a [feature request to map HTTP errors to PowerShell's error categories](https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/274).

`CmdletException` is the only AutoREST runtime exception event that's available for us to listen to, see https://github.com/Azure/autorest.powershell/blob/main/powershell/resources/psruntime/EventsExtensions.cs

Issue #901, opened 2021-10-13
New-MgEntitlementManagementAccessPackageAssignmentRequest StartDate timezone error


skyrahfall:
Hi,

I hope I do this right :-) first github issue ever.

When using the 
_New-MgEntitlementManagementAccessPackageAssignmentRequest_
cmdlet, the StartDateTime is not processed correctly,
leaving all created request in status "scheduled".

Run at 16:15 CEST /14:15 UTC
cmdlet sets 2021-10-13T14:15:42Z (**correct UTC timestamp**)
JSON in HTTP request:  "schedule": {"startDateTime": "2021-10-13T16:15:42.0000000+02:00"} (**local time again, marked as +2:00**)
JSON in HTTP response: "schedule":{"startDateTime":"2021-10-13T16:15:42Z" (**API has processed the local time as UTC**)

To me it looks like the API expects UTC, the code reflects this by converting the current time to UTC
```
    if ($RequestType -ne "AdminRemove") {
       if ($null -ne $StartDate -or $StartDate.Length -eq 0) {
            $now = Get-Date
            $ts = Get-Date $now.ToUniversalTime() -format "s"
            $StartDate = $ts + "Z"
        }
    }

```
side note: overwriting any existing value for StartDate which might be provided when calling the cmdlet.

Anyhow when the JSON is generated, the format is converted back to local time with the additional +2:00,
which is unfortunately ignored by the graph api.
This results in a local time being processed as UTC adding extra time (2h for my TZ).

The request gets processed 2h after running the powershell cmdlet.

Thanks

Here's some DEBUG output, which let to my assumption: 

> ﻿Scheduler was set: 2021-10-13T14:15:42Z
> DEBUG: CmdletBeginProcessing: 
> DEBUG: CmdletProcessRecordStart: 
> DEBUG: CmdletProcessRecordAsyncStart: 
> DEBUG: CmdletGetPipeline: 
> DEBUG: CmdletBeforeAPICall: 
> DEBUG: URLCreated: https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackageAssignmentRequests
> DEBUG: RequestCreated: https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackageAssignmentRequests
> DEBUG: HeaderParametersAdded: https://graph.microsoft.com/beta/identityGovernance/entitlementManagement/accessPackageAssignmentRequests
[...]

PR #989, opened 2021-12-21
New-MgEntitlementManagementAccessPackageAssignmentRequest ensure DateTime.Kind of StartDateTime is Utc

markwahl-msft:
In the Identity.Governance custom beta cmdlet, the cmdlet was setting the `StartDateTime` property by supplying a string value in UTC format.  However, issue #901 highlighted that this would override the caller's start date time, and also the generated output had the incorrect time zone.  So this PR addresses two problems.

First, for background, the underlying autogeneratied code contains a C# `DateTime` rather than `DateTimeOffset` for Edm.DateTimeOffset., Furthermore, the converters to JSON of a `DateTime` would use its `ToString()` method with a DateTimeFormat string ending in `K` (see for example  'date-time.ts' in autorest powershell for an example of that format string).  Unfortunately, the semantics of [the K format specifier](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings) is that the output string's time zone style varies depending on the `Kind` value inside the `DateTime`.  So when a caller, in PSh, sets a `DateTime` of a .NET object from a string, it is not safe to assume the automatic conversion PowerShell does from a customer's string results in UTC even if the string looks like it could be UTC.. This issue is not specific to the identity.governance module; the problem can occur elsewhere in msgraph-sdk-powershell, wherever a customer sets a `DateTime` from a string and then that `DateTime` is serialized into (for example) a GET filter parameter, or a POST, PUT or PATCH request payload.  When the `DateTime` has a `Kind` that is set to `Unspecified`, meaning that the `DateTime` parser earlier could not reliably guarantee that the Date and time the customer had supplied was in a UTC format, then the serialized JSON value from `toJSON()` is not valid OData because it is missing a time zone indicator, and that causes wierd errors to be returned by the Graph services. And if the `Kind` of the `DateTime` being serialized is `Local`, then the resulting JSON is not meeting the requirements of the Graph API, which state that time strings are in UTC, as waw seen in #901. 

This change removes setting the start date explicitly, and instead adds an check on the customer's supplied value, to force the `Kind` to always be `Utc` if the `Parse` method did not detect `Utc`.


peombwa (approved):
The change looks good!

I've also [opened a PR](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/997) that ensures string to date-time conversion is always in UTC, and all `DateTime` properties are serialized in UTC. This will be applied to all modules.

DateTimeOffset support will be added by AutoREST.PowerShell https://github.com/Azure/autorest.powershell/issues/584.

PR #1006, opened 2022-01-12
Add v1.0 Booking Business Under Solutions Singleton.

peombwa:
This PR adds v1.0 Bookings API under `solutions` singleton; see [API reference docs](https://docs.microsoft.com/en-us/graph/api/resources/booking-api-overview?view=graph-rest-1.0#using-the-bookings-rest-api).

Open Question(s):
- [x] Should we drop `solution` from v1.0 command names so that they match beta commands? `Get-MgSolutionBookingBusiness` ([v1.0](https://docs.microsoft.com/en-us/graph/api/bookingbusiness-list?view=graph-rest-1.0#request-1)) vs `Get-MgBookingBusiness` ([beta](https://docs.microsoft.com/en-us/graph/api/bookingbusiness-list?view=graph-rest-beta&tabs=http#request-1)).

@darrelmiller, would you know if there are plans to ensure v1.0 and beta Bookings APIs use the same singleton/entitySet?

cc\ @maisarissi


maisarissi:
I see implications that could affect users for both keeping names different or the same.
Having different names could imply there are 2 different commands, hitting different APIs or returning different responses. It could also prevent users from finding the corresponding commands in v1 and beta.
Now, renaming v1 to match beta will solve the problems above, but since the API owners decided to drop `solution` this could be a hint they want users to now something changed, like the responses or something else, or could be only refactoring/cleaning.
I would argue that if there is no such change, the responses are kind of the same and how the API deal with the request is the same/similar, we can drop `solution` to improve discoverability and user experience.

peombwa:
@maisarissi, that's my thinking as well. I should mention that the API has been available in beta for a few years, without solutions. Bookings API recently (12/2021) moved to v1.0 under `solutions` singleton, but the change didn't make it to beta.

maisarissi:
I would bet on renaming it then 😄

peombwa:
@georgend, @timayabi2020 Can I get a review on this PR.

georgend (approved):

georgend (approved):

PR #1029, opened 2022-01-25
Docs Pipeline Git Permission Issue

georgend:
Git Permissions Issue causing docs generation from producing file artefacts. 
Remove recurse step since it can be set at the pipeline level and it causes errors. 
Use Run CI parameter. 
Results Confirmation:-https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/1048


maisarissi (approved):

.azure-pipelines/common-templates/checkout.yml
```diff
@@ -32,15 +32,7 @@ steps:
...
         git status
 
-  - task: Bash@3
-    displayName: "Update Git Submodules"
-    inputs:
-      targetType: inline
-      script: |
-        git pull --recurse-submodules
-        git submodule update --init --recursive --remote
-
```
peombwa, line 42: This change is breaking weekly OpenAPI refreshes. We need to update the AutoREST submodule using from the remote before we use it; otherwise, you'll get this error: https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=68036&view=logs&j=cf6261d7-9a9d-56e6-29df-71d75249ff94&t=5e28baed-0a38-5f5d-2604-06f99369333d.

I'll revert this change.

PR #1067, opened 2022-02-09
Agent Change

georgend:


peombwa:
@finsharp, are we changing the building agent to MsGraphBuildAgentsWindowsPsSDK? What necessitated the change?

PR #1179, opened 2022-03-29
Fixes Missing `-Debug` Request Body in PS 5.1

### LINKED ISSUE #980: Body of request is not shown in debug
Expected: Body would show the JSONified request for the new group.
Actual: Nothing is shown. I have to go to fiddler or similar capture mechanism to see it.

```
❯ $x | new-mggroup -verbose -debug
DEBUG: [CmdletBeginProcessing]: - New-MgGroup begin processing with parameterSet 'Create'.
DEBUG: [Authentication]: - AuthType: 'Delegated', AuthProviderType: 'InteractiveAuthenticationProvider', ContextScope: 'CurrentUser', AppName: 'Microsoft Graph PowerShell'.
DEBUG: [Authentication]: - Scopes: [Directory.AccessAsUser.All, EntitlementManagement.Read.All, EntitlementManagement.ReadWrite.All, openid, PrivilegedAccess.ReadWrite.AzureAD, PrivilegedAccess.ReadWrite.AzureADGroup, PrivilegedAccess.ReadWrite.AzureResources, profile, User.Read, email].

Confirm
Are you sure you want to perform this action?
Performing the operation "New-MgGroup_Create" on target "Call remote 'GroupsGroupCreateGroup' operation".
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"): y
DEBUG: ============================ HTTP REQUEST ============================

HTTP Method:
POST

Absolute Uri:
https://graph.microsoft.com/v1.0/groups

Headers:
SdkVersion                    : Graph-powershell-1.9.0,Graph-dotnet-1.25.1
FeatureFlag                   : 00000047
Cache-Control                 : no-store, no-cache
User-Agent                    : Mozilla/5.0,(Windows NT 10.0; Microsoft Windows 10.0.22000; en-US),PowerShell/7.2.0,New-MgGroup_Create
Accept-Encoding               : gzip

Body:


```

peombwa:
This PR fixes #980 by falling back to `OnBeforeCall` event handler when `-Debug` is used in PS 5.1.

#### Root cause of the bug
In PS 5.1 (NET Framework 4.x), the requestMessage content was being disposed before we could access it in the `OnResponseCreated` event handler. We need to log the request object in the `OnResponseCreated` to retrieve the request headers set via the middleware pipeline.

#### Fix
1. In PS 5.1 (.NET Framework), log the request message object in the `OnBeforeCall` event handler.
2. In PS 6+ (NET Core, NET 5+), log the request message object in the `OnResponseCreated` event handle.

### DIFF
--- a/tools/Custom/Module.cs
+++ b/tools/Custom/Module.cs
@@ -62,6 +62,10 @@ namespace Microsoft.Graph.PowerShell
                 case Events.CmdletBeginProcessing:
                     await OnCmdletBeginProcessing(id, cancellationToken, getEventData, signal, invocationInfo);
                     break;
+                case Events.BeforeCall:
+                    if (IsNetFramework())
+                        await OnBeforeCall(id, cancellationToken, getEventData, signal);
+                    break;
                 case Events.ResponseCreated:
                     await OnResponseCreated(id, cancellationToken, getEventData, signal);
                     break;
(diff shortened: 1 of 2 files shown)


timayabi2020 (approved):

tools/Custom/Module.cs
```diff
@@ -149,5 +170,10 @@ private async Task OnCmdletEndProcessing(string id, CancellationToken cancellati
                 }
             }
         }
+
+        private bool IsNetFramework()
```
georgend, line 174: Is there a better way to detect that we are running in NETFramework?

peombwa: `RuntimeInformation.FrameworkDescription` should be enough to detect the name of the .NET installation an app is running on. See https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=netframework-4.7.2.

I'll change the implementation to check the PS host version instead since we are already capturing it in `(Get-MgContext).PSHostVersion`.

timayabi2020 (approved):

Issue #1144, opened 2022-03-11
Replace Deprecated Auth Providers with Azure Identity


peombwa:

peombwa:
As part of this work item, we should also improve our test coverage by adding more pester tests.

peombwa:
Closed by https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/1201

PR #1184, opened 2022-03-31
Reuse HTTP Client In Session

### LINKED ISSUE #1108: Excessive TCP Network Connections
I've come across an issue where running a Powershell script in Azure runbook sandbox errors with 'reached limit 1000 network sockets'.
The script reads all applications in my Azure AD Tenant (over 3500) and then writes some details to Sharepoint lists.
When calling New-MgSiteListItem or Update-MgSiteListItem. each call creates a new connection to Sharepoint. Within a few minutes the 1000 limit is reached.
In my Dev environment and using TCPView you can see the 'established' connections rise to over 1000 within minutes.

When calling these functions why are new connections established on each call? 
Is there any way to mitigate this?

Thanks

peombwa:
This PR fixes #1108 by ensuring we initialize the HTTP client only once per session and reuse it for subsequent calls to the service. The PR also adds unit tests to validate the change on NET4x and NETCore.

The following charts show the network connection count trend (`netstat -an`) when you run `Get-MgSecuritySecureScore` 50 times.
#### TCP Connection Count in PS 5.1 (before the fix)
image

#### TCP Connection Count in PS 5.1 (after the fix)
image

#### Notes:
From my testing, PS 7 automatically closes the connection after each request. Thus, the socket exhaustion issue is only present in PS 5.1 and Azure Automation Runbook, which runs on PS 5.1 and has a max socket limit of 1000.
image


src/Authentication/Authentication.Test/Helpers/HttpHelpersTests.cs
```diff
@@ -71,5 +72,112 @@ public void GetGraphHttpClientShouldReturnHttpClientWithCustomerProvidedTimeout(
...
+            Assert.NotNull(httpClient);
+            Assert.NotNull(GraphSession.Instance.GraphHttpClient);
+            Assert.Equal(httpClient.BaseAddress, GraphSession.Instance.GraphHttpClient.BaseAddress);
+
+            // reset static instance.
+            GraphSession.Reset();
+        }
+
+        [Fact]
+        public void GetGraphHttpClientShouldReturnNewHttpClientSignOutThenSignIn()
```
ddyett, line 142: shouldn't there be a test that win disconnect-mggraph is done the GraphHttpClient instance is null.

timayabi2020 (approved):

src/Authentication/Authentication.Test/Helpers/HttpHelpersTests.cs
peombwa: I'm already mocking disconnect-mggraph in https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/1184/files/adb98883d4da7d3925cad95307e0e36f4ff9dbac#diff-33c29c1e330c4609e494c23e4a071b20d7e67bedd59120cd21d2511b69fe08c3R159-R161, by explicitly setting AuthContext and GraphHTTPClient to null. We don't have a way of invoking disconnect-mggraph directly in an Xunit test.

peombwa: I'll introduce pester tests in https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1144 to cover invoking of commands in a powershell process.

PR #1189, opened 2022-04-04
Identity governance custom cmdlets: ensure existing cmdlets with the same names are only loaded in beta profile (for now)

markwahl-msft:

In the identity governance Graph API for entitlement management, some resource type paths are different between beta and v1.0, and some structures are different between beta and v1.0.

In the identity governance module, some of the custom cmdlets have names that correspond to their Graph paths.  As a result, some are for beta only, for instance `Get-MgEntitlementManagementAccessPackageAssignment.ps1` and others for v1.0 only, for instance `Get-MgEntitlementManagementAssignment.ps1`, set with the `Microsoft.Graph.PowerShell.Profile`.   

For the `AccessPackage` and `ConnectedOrganization` resource types, the names of the Graph paths are the same between beta and v1.0, but the resource type structures generated from Graph for those two are different, with different field and relationship names.  Currently there are four custom cmdlets that were written originally for beta, but because they don't have a `Microsoft.Graph.PowerShell.Profile`, those appear in the 1.9.3 v1.0 module for both v1.0 and beta.  Unfortunately, when they're loaded for v1.0, they call beta Graph APIs and return the beta version of the structures. In order to prevent beta APIs from being inadvertently called and confusing the caller, this PR temporarily changes those four cmdlets to only be in beta profile.

A later PR for these would allow those four cmdlets to also be used in v1.0 with the v1.0 resource type structures, that PR would be substantially bigger than this one as it would need to call different private APIs internally, and have different output types depending on whether the beta or v1.0 profile was selected..


peombwa (approved):

markwahl-msft:
#sign-off

PR #1190, opened 2022-04-04
Identity governance: suppress Invoke-MgFilterEntitlementManagementCatalogAccessPackage

markwahl-msft:

In the Azure AD identity governance module, Invoke-Mg cmdlets are generated for the special-purpose filtering Microsoft Graph APIs for entitlement management, such as an access package or a catalog filterByCurrentUser.  This PR is intended to suppress the automatic generation of one cmdlet in that module. 

The URL that the `Invoke-MgFilterEntitlementManagementCatalogAccessPackage` cmdlet sends is a URL that is supported by entitlement management as part of its API surface.

(The other `Invoke-Mg*EntitlementManagement*` cmdlets for entitlement management should be unchanged by this PR. While those other ones _are_ sending valid requests that are processed by the service, the cmdlets still fail when parsing the _responses_ at present, because they don't realize that the OData response is a structure containing an array (and potentially a nextLink), rather than an array.   That's the subject of a separate existing bug.)


peombwa:
@markwahl-msft, should we also remove `ContainsTarget="true"` from `accessPackages` NavigationProperty in `accessPackageCatalog` entity if `/identityGovernance/entitlementManagement/catalogs/{accessPackageCatalog-id}/accessPackages*` is not supported by the service? `ContainsTarget=true` is what we rely on to expand the paths:
``` xml
<EntityType Name="accessPackageCatalog" BaseType="graph.entity">
  <Property Name="catalogType" Type="graph.accessPackageCatalogType" />
  <Property Name="createdDateTime" Type="Edm.DateTimeOffset" />
  <Property Name="description" Type="Edm.String" />
  <Property Name="displayName" Type="Edm.String" />
  <Property Name="isExternallyVisible" Type="Edm.Boolean" />
  <Property Name="modifiedDateTime" Type="Edm.DateTimeOffset" />
  <Property Name="state" Type="graph.accessPackageCatalogState" />
  <NavigationProperty Name="accessPackages" Type="Collection(graph.accessPackage)" ContainsTarget="true" />
</EntityType>
```

peombwa (approved):

markwahl-msft:
> ContainsTarget=true is what we rely on to expand the paths

That is a good question.  One of the challenges is that today Graph is using this for two semantically distinct concepts - logical containment relationship of one type of resource to another, and URL navigation between those resources.   Perhaps in the future there will be different tagging properties for these two different cases.

markwahl-msft:
#sign-off

Issue #1002, opened 2022-01-07
Investigate Dependabot Integration That Works with Assembly Load Context


peombwa:
This is part of #998. We need to ensure dependencies in the Assembly Load Context are always updated when dependabot detects updates.

References:
- [PowerShell assembly dependency conflicts](https://docs.microsoft.com/en-us/powershell/scripting/dev-cross-plat/resolving-dependency-conflicts?view=powershell-7.2)
- [Our dependency resolver](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/src/Authentication/Authentication/Utilities/DependencyAssemblyResolver.cs#L14)

ddyett:
not feasible to use dependabot with powershell model.

PR #1201, opened 2022-04-12
Move to Azure Identity

### LINKED ISSUE #1144: Replace Deprecated Auth Providers with Azure Identity

peombwa:
This PR closes #1144 by:
- Replacing deprecated auth providers with Azure Identity's token credentials.
- Updating both unit and Pester tests.

The following improvements have been made to the Authentication module as part of this transition:
- Drops `-AccessToken` from `Invoke-MgGraphRequest`. The goal is to have one authentication/authorization point, `Connect-MgGraph`.
- Drops `-ForceRefresh` from `Connect-MgGraph`. `-ForeceRefresh` is not exposed by Azure Identity as the library will fetch new tokens automatically when new scopes are added.
- Adds a simplified `In-Memory` token cache for process context scope.
- Updates dependencies.


timayabi2020 (approved):

ddyett:
do you recall why we added the accesstoken on invoke-mggraphrequest to begin with?

src/Authentication/Authentication.Core/Microsoft.Graph.Authentication.Core.csproj
```diff
@@ -1,22 +1,27 @@
...
-    <PackageReference Include="Microsoft.Graph.Core" Version="1.25.1" />
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.29.0" />
-    <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" Version="5.6.0" />
-    <PackageReference Include="Microsoft.IdentityModel.Logging" Version="5.6.0" />
-    <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="5.6.0" />
-    <PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.19041.1" />
-    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.6.0" />
+    <PackageReference Include="Azure.Core" Version="1.22.0" />
+    <PackageReference Include="Azure.Identity" Version="1.5.0" />
+    <PackageReference Include="Microsoft.Graph.Core" Version="2.0.8" />
```
ddyett, line 15: by the time this goes out wouldn't kiota likely be the release?  Is the plan to convert this later and align with CLI?

ddyett:
can you say more about why the token caches for mac/linux are no longer necessary?

src/Authentication/Authentication.Core/TokenCache/InMemoryTokenCacheOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Graph.PowerShell.Authentication.Core.TokenCache
+{
+    public class InMemoryTokenCacheOptions : UnsafeTokenCacheOptions
+    {
+        private readonly ReaderWriterLockSlim _sessionLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
+        internal ReadOnlyMemory<byte> TokenCache { get; private set; }
```
ddyett, line 16: seems odd that the options class is holding the reference to the cache and operating on the object.

peombwa: Good point! It is a cache manager. It's Azure Identity that named the base class [UnsafeTokenCacheOptions](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.unsafetokencacheoptions?view=azure-dotnet). I opted to use the options suffix to be in sync with Azure Identity naming.

src/Authentication/Authentication.Core/Microsoft.Graph.Authentication.Core.csproj
peombwa: That's the plan. Swapping the current MS Graph .NET core for Kiota based core should be trivial. I have a work item to integrate dependabot for verion updates - https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1002.

peombwa:
> do you recall why we added the accesstoken on invoke-mggraphrequest to begin with?

It was because we didn't expose `-AccessToken` on `Connect-MgGraph` at the time.

peombwa:
> can you say more about why the token caches for mac/linux are no longer necessary?

We no longer need them because Azure.Identity securely caches tokens on Windows, MacOS, and Linux (when libsecret-1.so.0 is available). We will fall back to in-memory cache when Azure.Identity can't persist tokens to the underlying OS mechanism ([MsalCachePersistenceException](https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/master/src/Microsoft.Identity.Client.Extensions.Msal/MsalCachePersistenceException.cs) is thrown).

peombwa:
We should verify that successfully sign in and token caching is possible on the following platforms (I'll check them once verified):
- Windows:
  - [ ] PS 5.1
  - [ ] PS 7
  - [ ] WSL
- MacOS:
  - [ ] Native PS core session via terminal
  - [ ] Docker
- Linux:
  - [ ] Cloud Shell
  - [ ] Ubuntu Server

georgend:
> do you recall why we added the accesstoken on invoke-mggraphrequest to begin with?

We had added it for BYOT scenarios. For example authenticating against MSI, getting a token and (re)-using it with the SDK.

georgend (approved):

Issue #1224, opened 2022-04-20
Get-MgServiceAnnouncementMessage


timayabi2020:
Remove default page size of 100 from ```
Get-MgServiceAnnouncementMessage -All``` command and work with the number of items returned by the service.
The goal is to get the "odata.nextLink" from the response.

peombwa:
@timayabi2020, we will need to review the current paging logic as there is a recursion when calling `onOk`->`API_Call`->`onOk`.

See https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1192.

Issue #1141, opened 2022-03-11
"Get-MgServiceAnnouncementMessage -All" only loads first page


buckbanzai:
Hello,
When running the following command, it only outputs 100 records, where I would expect it to page through all available results (259 in this case).
```
$ServiceAnnouncementMessages = Get-MgServiceAnnouncementMessage -All -CountVariable 'ServiceAnnouncementMessageCount'
PS /Users/liam.steckler> $ServiceAnnouncementMessages.Count
100
PS /Users/liam.steckler> $ServiceAnnouncementMessageCount
259
```

SDK version 1.9.3, running on PowerShell Core 7.2.1 on macOS

timayabi2020:
> Hello, When running the following command, it only outputs 100 records, where I would expect it to page through all available results (259 in this case).
> 
> ```
> $ServiceAnnouncementMessages = Get-MgServiceAnnouncementMessage -All -CountVariable 'ServiceAnnouncementMessageCount'
> PS /Users/liam.steckler> $ServiceAnnouncementMessages.Count
> 100
> PS /Users/liam.steckler> $ServiceAnnouncementMessageCount
> 259
> ```
> 
> SDK version 1.9.3, running on PowerShell Core 7.2.1 on macOS
@buckbanzai This is both a service and an SDK issue.  Whenever you provide the argument "-All" the SDK sends a request with a default page size of 100 (https://graph.microsoft.com/v1.0/admin/serviceAnnouncement/messages?$top=100), expecting next link to be provided ( "@odata.nextLink": "https://graph.microsoft.com/v1.0/admin/serviceAnnouncement/messages?$Select=id&$skip=100"). Unfortunately this is not happening, therefore a service issue has been raised to address this problem. 
Meanwhile on the sdk side we will remove the default page size so that we just work with the size we are getting from the service.
Below is a screenshot that summarizes the above explanation.
**Next link provided without specifying a page size**
```Get-MgServiceAnnouncementMessage -CountVariable 'ServiceAnnouncementMessageCount' -Debug```
![Screenshot 2022-04-05 172941](https://user-images.githubusercontent.com/10947120/161793633-275ce329-c52a-4b9b-82ce-978ef864e2cd.png)
**Page size specified yet there is no next link**
```Get-MgServiceAnnouncementMessage -All -CountVariable 'ServiceAnnouncementMessageCount' -Debug```
[...]

PR #1230, opened 2022-04-26
Set the default page size to MaxPageSize

timayabi2020:
The default page size has been set to the maximum value (999) as defined by the constant variable "MaxPageSize" whenever the argument "All" is provided. E.g. 
```
Get-MgServiceAnnouncementMessage -All -CountVariable 'ServiceAnnouncementMessageCount' -Debug
```
This is supposed to address issues https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1141 and https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1224.
The screenshot shows the result of this fix.  i.e., ```odata.nextLink``` parameter which was not there before as per the customer complaint, is now available.
image

### DIFF
--- a/tools/Custom/ListCmdlet.cs
+++ b/tools/Custom/ListCmdlet.cs
@@ -107,7 +107,17 @@ namespace Microsoft.Graph.PowerShell.Cmdlets.Custom
                 limit = top;
             }
 
-            int currentPageSize = invocationInfo.BoundParameters.ContainsKey("PageSize") ? PageSize : DefaultPageSize;
+            int currentPageSize = DefaultPageSize;
+            if(invocationInfo.BoundParameters.ContainsKey("PageSize"))
+            {
+                currentPageSize = PageSize;
+            }
+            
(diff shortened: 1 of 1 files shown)


peombwa:
@timayabi2020, please test this assumption with other workloads, as each workload has a different paging strategy. For example, in AzureAD (users, groups, devices, etc.), the default is page size 100 and max 999. In Outlook, the default page size is 10, and the max is a black box.
> - Different APIs might have different default and maximum page sizes.
> - Different APIs might behave differently if you specify a page size (via the $top query parameter) that exceeds the maximum page size for that API. Depending on the API, the requested page size might be ignored, it might default to the maximum page size for that API, or Microsoft Graph might return an error.

See https://docs.microsoft.com/en-us/graph/paging?view=graph-rest-1.0

peombwa:
@timayabi2020, the implementation will not work for some Team's commands/APIs such as `Get-MgChat -All -PageSize 999`. See https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1248 for more details.

Issue #1312, opened 2022-06-08
Invoke-MgGraphRequest resets Environment


remyloy:
After a call to Invoke-MgGraphRequest the selected environment through Connect-MgGraph is not restored.
The following snippet shows this issue by adding a test environment, which is configured the same as the built-in Global environment.

Just for clarification, this is a simplification for demonstrating the issue. In my production code the environment uses a different graph endpoint.

```powershell
#Requires -Modules @{ModuleName = "Microsoft.Graph.Authentication"; ModuleVersion = "1.9.6" }

# Connection parameters
$Params = @{
    ClientId              = ''
    TenantId              = ''
    CertificateThumbprint = ''
}
$TestEnv = Add-MgEnvironment -Name 'Test' -AzureADEndpoint 'https://login.microsoftonline.com' -GraphEndpoint 'https://graph.microsoft.com'
try
{
    Connect-MgGraph @Params -Environment $TestEnv.Name | Out-Null
    if ($null -eq [Microsoft.Graph.PowerShell.Authentication.GraphSession]::Instance.Environment)
    {
        Write-Error -Message 'Env not set (before Invoke-MgGraphRequest) '
    }

    Invoke-MgGraphRequest -Uri '/v1.0/users' | Out-Null

    if ($null -eq [Microsoft.Graph.PowerShell.Authentication.GraphSession]::Instance.Environment)
    {
        Write-Error -Message 'Env not set (after Invoke-MgGraphRequest)'
    }
}
finally
{
    Disconnect-MgGraph
    Remove-MgEnvironment -Name $TestEnv.Name | Out-Null
}
```

The error is written after Invoke-MgGraphRequest, as the environment is then null instead of being the Test environment.

My current workaround is to call Connect-MgGraph, before each call to Invoke-MgGraphRequest, to ensure the expected environment is actually used.

PR #1328, opened 2022-06-20
Removed null reference to Graph Instance environment after calling Invoke-MgGraph command

timayabi2020:
﻿

Fixes #

- Reassign _originalEnvironment variable to the previous user defined environment instead of pointing to null. This will fix environment null reference issues on Invoke-MgGraph command such as this https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1312

The images below show the results before and after proposed fix respectively

- **Before fix**
      
image

- **After fix**
     
image

### DIFF
--- a/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
+++ b/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
@@ -1090,9 +1090,13 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
 
         /// <summary>
         ///     Resets GraphSession environment back to its original state.
+        ///     Original state can remain to be the previous state defined by the user, 
+        ///     or the default global state. User defined state can be removed by calling,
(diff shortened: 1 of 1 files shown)


src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
```diff
@@ -1090,9 +1090,13 @@ private string QualifyFilePath(string path)
 
         /// <summary>
         ///     Resets GraphSession environment back to its original state.
+        ///     Original state can remain to be the previous state defined by the user, 
+        ///     or the default global state. User defined state can be removed by calling,
+        ///     Remove-MgGraph Command
```
peombwa, line 1095: We don't have a command called Remove-MgGraph. Did you mean `Disconnect-MgGraph`?

timayabi2020: @peombwa I meant Remove-MgEnvironment

PR #1379, opened 2022-07-09
Add NewMgGroupOwner

DanielWahlgren:
﻿

- Adds New-MgGroupOwners to ensure members and owners can be added in ways closely related.

### Other information
I made this primarily to challenge myself. I use Graph a lot and like many write scripts directly agains MS Graph, not using the SDK since it is fairly clunky. 
I will consider making more if this is successful, aswell as opening issues for things that are outside of my abilities to fix myself.


peombwa (approved):
Thank you for your contribution!

The change looks good to me! Users of this module will find this useful!!

timayabi2020 (approved):
I have played around with it and all is well. Thanks

PR #1387, opened 2022-07-11
Current User and Local Machine Certificate Store Lookup

### LINKED ISSUE #936: Allow localmachine store search with GetCertificateByThumbprint and GetCertificateByName
Hello graph sdk PowerShell team, My name is Monte - I'm an internal MSFT employee.  This came up in a discussion with a customer, so relaying here. 

**Problem statement**

When authenticating a service principal inside of an Azure automation hybrid runbook worker, service principal certificates for azure automation are exported to the local machine certificate store.  This leads to inconsistent behavior when authenticating to graph in a cloud worker vs a hybrid worker since Connect-MgGraph only looks in the current user store during authentication.  

**Workaround**

Customers using a hybrid runbook worker have to find their own certificate and pass the certificate object to connect-mggraph via the certificate parameter in their runbooks.

**Solution**

By allowing the [GetCertificateByThumbprint](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/036a9c064815bd7074cb846e858456b4cc51622d/src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs#L214) and [GetCertificatebyName ](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/036a9c064815bd7074cb846e858456b4cc51622d/src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs#L244)methods to look in the Local Machine store, we can alleviate users from having to manually find certificates on their HRW to successfully auth to graph.

peombwa:
This PR closes #936 by providing a certificate lookup implementation that looks for certificates in both the current user and local machine certificate store.

### DIFF
--- a/src/Authentication/Authentication.Core/ErrorConstants.cs
+++ b/src/Authentication/Authentication.Core/ErrorConstants.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Graph.PowerShell.Authentication.Core
             internal const string AuthenticationTimeout = "Authentication timed out after {0} seconds due to inactivity. Please try again.";
             internal const string InvalidUserProvidedToken = "The provided access token is invalid. Set a valid access token to `-{0}` parameter and try again.";
             internal const string ExpiredUserProvidedToken = "The provided access token has expired. Set a valid access token to `-{0}` parameter and try again.";
+            internal const string CertificateNotFound = "Certificate with {0} '{1}' was not found in certificate store or has expired.";
         }
     }
 }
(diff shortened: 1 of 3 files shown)


timayabi2020:
> This PR closes #936 by providing a certificate lookup implementation that looks for certificates in both the current user and local machine certificate store.

Is their a possibility of

> This PR closes #936 by providing a certificate lookup implementation that looks for certificates in both the current user and local machine certificate store.

Does this still cater for those who would prefer to have the certificates somewhere else instead of the default local machine certificate store? I'm suspecting there are some customers who would still prefer to pass the certificate object

peombwa:
@timayabi2020, `-Certificate` is currently supported - https://github.com/microsoftgraph/msgraph-sdk-powershell#usage. This PR doesn't change how `-Certificate` works.

timayabi2020 (approved):

PR #1402, opened 2022-07-18
Microsoft Graph Authentication parameter completers

### LINKED ISSUE #1386: Connect-MgGraph: Environment parameter does not support auto-complete
The `Environment` parameter of `Connect-MgGraph` requires the user to use `Get-MgEnvironment` to enumerate environments. Instead, we should add a parameter completer that does this so the user gets auto-complete.

We should also add this completer to any other commands in `Microsoft.Graph.Authentication` that take `Environment` as a parameter, such as `Get-MgEnvironment`.

FehintolaObafemi:
﻿

Fixes #1386 

- Adding parameter completers to Connect-MgGraph parameters

### Other links

- #1386

### DIFF
--- a/src/Authentication/Authentication/custom/ParameterCompleters.ps1
+++ b/src/Authentication/Authentication/custom/ParameterCompleters.ps1
@@ -9,7 +9,7 @@ Set-StrictMode -Version 2
 $scopesParameterBlock = {
     param($commandName,
     $parameterName,
-    $wordToComplete,
+    $wordToCompleteScopes,
     $commandAst,
     $fakeBoundParameters
     )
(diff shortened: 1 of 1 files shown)


src/Authentication/Authentication/custom/ParameterCompleters.ps1
```diff
@@ -18,8 +18,25 @@ $scopesParameterBlock = {
...
+$environmentParameterBlock = {
+    param($commandName,
+    $parameterName,
+    $wordToCompleteEnvironment,
+    $commandAst,
+    $fakeBoundParameters
+    )
+
+    $permissions = Get-MgEnvironment | Select-Object -ExpandProperty Name
+    $permissions | Where-Object {
```
peombwa, line 34:
```suggestion
    $environments = Get-MgEnvironment | Select-Object -ExpandProperty Name
    $environments | Where-Object {
```

peombwa (approved):

PR #1484, opened 2022-08-31
Pester tests assertion update

timayabi2020:
﻿

- Update pester test assertion for module imports that write to streams when debug preference is set to 'Inquire'. The streams debug count returns a value of 2 yet the expected value is 0.
image

### DIFF
--- a/tools/TestModule.ps1
+++ b/tools/TestModule.ps1
@@ -16,7 +16,6 @@ if(-not $Isolated) {
 }
 
 $modulePsd1 = Get-Item -Path (Join-Path $ModulePath "./$ModuleName.psd1")
-Write-Host $modulePsd1
 $LocalLoadEnvPS1 = Join-Path $PSScriptRoot 'Tests/loadEnv.ps1'
 $AuthModulePSd1 = Join-Path $PSScriptRoot "../src/Authentication/Authentication/artifacts/Microsoft.Graph.Authentication.psd1"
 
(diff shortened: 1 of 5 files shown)


tools/TestModule.ps1
```diff
@@ -35,6 +36,6 @@ $PesterConfiguration.TestResult.Enabled = $true
 $PesterConfiguration.TestResult.OutputPath = (Join-Path $ModuleTestsPath "$moduleName-TestResults.xml")
 
 $TestResults = Invoke-Pester -Configuration $PesterConfiguration
-If ($TestResults.FailedCount -gt 0) { Write-Error "$($TestResults.FailedCount) tests failed." }
```
peombwa, line 38: @timayabi2020, why are we changing this to 5? We shouldn't have filing tests.

tools/TestModule.ps1
```diff
@@ -16,6 +16,7 @@ if(-not $Isolated) {
 }
 
 $modulePsd1 = Get-Item -Path (Join-Path $ModulePath "./$ModuleName.psd1")
+Write-Host $modulePsd1
```
peombwa, line 19: Let's remove the log.

tools/TestModule.ps1
timayabi2020: Thank you for flagging this. It was an error from my side. This has been changed

tools/TestModule.ps1
timayabi2020: This has been removed

peombwa (approved):

PR #1555, opened 2022-10-05
Adds Client Secret Credentials Support

### LINKED ISSUE #1366: Add -Credential as authentification method to the Connect-MgGraph Command
Hello,
I would like to request adding -Credential as authentification method to the Connect-MgGraph Command.
This is needed to use the Microsoft Graph PowerShell SDK in DevOps. With the other ways there are following Problems:

 - Window opening to log in: 
Not possible in DevOps, as windows can't open there

 - Device code: 
very complicated as you don't know when you need to do it especially with longer scripts and you can miss the time slot to do it really often
 
 - Application with certificate: 
The scripts should be useable for many tenants and most of the time they only need to run once. So for another script where I use MS Graph Headers for direct command authentification I alread have a script part which creates and afterwards deletes the application, whats possible with only admin credentials, while it's not possible to just use the admin credentials to authentificate.
But when I want to use Microsoft Graph PowerShell SDK cmdlets I can't do it as I need the certificate on top. Any authentification with certificate isn't possible in DevOps as I would need a certificate of the computer where I run the script, but the script runs on a unkown DevOps Server.

peombwa:
This PR closes #1366 by adding `-Credential` parameter of type [PSCredential](https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/add-credentials-to-powershell-functions?view=powershell-7.2#creating-credential-object) to `Connect-MgGraph` command.

The parameter is being added to:
- Support the authentication scenarios mentioned in #1366.
- Align with MS Graph CLI as mentioned in https://github.com/microsoftgraph/msgraph-cli/issues/147#issuecomment-1224610883.

### Proposes Usage
image


timayabi2020 (approved):

src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
```diff
@@ -180,12 +188,14 @@ private async Task ProcessRecordAsync()
...
                         }
                         break;
-                    case Constants.AccessTokenParameterSet:
+                    case Constants.AppSecretCredentialParameterSet:
                         {
-                            authContext.AuthType = AuthenticationType.UserProvidedAccessToken;
-                            authContext.TokenCredentialType = TokenCredentialType.UserProvidedAccessToken;
-                            authContext.ContextScope = ContextScope.Process;
-                            GraphSession.Instance.InMemoryTokenCache = new InMemoryTokenCache(Encoding.UTF8.GetBytes(new NetworkCredential(string.Empty, AccessToken).Password));
+                            authContext.AuthType = AuthenticationType.AppOnly;
```
nkasco, line 193: Won't this cause user confusion when users inevitably try to use `-Credential` within `Connect-MgGraph` for ROPC / User Credential auth flow use with delegated permission types? I get that you're only intending to add support for client secrets / App Only, I just think in the process it might be creating UX confusion. 

Perhaps an alias or rename of the parameter could aleviate this. `-ClientSecret`, `-ClientSecretCredential`, `-AppCredential`, or `-AppSecret` might be a few possibilities.

grzegorzwierzbicki:
This is completely wrong and confusing
-Credential should accept username and password for accessing the application with delegated permissions.

grzegorzwierzbicki:
On top of that, the v2 module no longer accepts tokens acquired with MSAL.PS
Why are you doing this ?
Why are you working so hard on making life miserable for anyone who just wants to use Graph in a script.

```
$AccessToken = Get-MsalToken -ClientId $PowerShellGraphAppID -TenantId $TenantId -UserCredential $creds -Scopes "User.ReadWrite.All"
Connect-MgGraph -AccessToken $AccessToken.AccessToken
Connect-MgGraph : Cannot bind parameter 'AccessToken'. Cannot convert the "eyJ....5bA" value of type "System.String" to type "System.Security.SecureString"
```

nkasco:
> On top of that, the v2 module no longer accepts tokens acquired with MSAL.PS
> 
> Why are you doing this ?
> 
> Why are you working so hard on making life miserable for anyone who just wants to use Graph in a script.
> 
> 
> 
> ```
> 
> $AccessToken = Get-MsalToken -ClientId $PowerShellGraphAppID -TenantId $TenantId -UserCredential $creds -Scopes "User.ReadWrite.All"
> 
> Connect-MgGraph -AccessToken $AccessToken.AccessToken
> 
> Connect-MgGraph : Cannot bind parameter 'AccessToken'. Cannot convert the "eyJ....5bA" value of type "System.String" to type "System.Security.SecureString"
> 
> ```

You need to convert the string token to secure string to use with the v2 module. That was documented in the breaking changes. MSAL.PS isn't broken. 

That said, I do agree about the credentials parameter being confusing as I documented above.

PR #1739, opened 2023-01-10
Fix bad https link

jbzdarkid:
I'm not sure if this markdown is auto-generated, if so, this should be fixed at the source.
As-is, this link is clickable in the github rendered UI, except that github "helpfully" assumes that the period is part of the link.
Since this isn't really a sentence, I don't think the punctuation is helping anyone -- so I'm proposing that you remove it.

### DIFF
--- a/src/Identity.SignIns/Identity.SignIns/readme.md
+++ b/src/Identity.SignIns/Identity.SignIns/readme.md
@@ -73,5 +73,5 @@ directive:
 
 ``` yaml
 module-version: 1.20.0
-release-notes: See https://aka.ms/GraphPowerShell-Release.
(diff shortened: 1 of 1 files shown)


peombwa:
Thanks for your contribution!

We no longer set the link in the readme in v2.x - https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/features/2.0/src/Identity.SignIns/Identity.SignIns.md.

Yes, the file is autogenerated in v1.x and is not changeable at this point. The link is only ever used to generate the release notes link in PowerShell Gallery and is not meant to be clickable in GitHub. The versioning section is formatted as yaml.

Closing as this is no longer needed in v2.x.

PR #1753, opened 2023-01-20
Teams examples migration

### LINKED ISSUE #1559: Provide tooling to migrate examples to v2
We should provide tooling to migrate examples to v2. The tool should be built in such a way that customers can use it to migrate from v1 to v2.

The tool should:
- [ ] Rename existing beta example file names to match new beta command names, i.e., VERB-MgBetaNOUN.md. See [msgraph-sdk-powershell/src/Users/Users/examples/v1.0-beta at dev · microsoftgraph/msgraph-sdk-powershell (github.com)](https://github.com/microsoftgraph/msgraph-sdk-powershell/tree/dev/src/Users/Users/examples/v1.0-beta).
- [ ] Find and replace commands in the examples to match new beta command names.

This task will involve:
1.  Generating modules locally while on the features branch
2.  Checking first if example files exist on the feature branch before copying v1.0 and beta example files from dev branch
3.  Conforming the cmdlets to the naming convention as per the design [design doc](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/features/2.0/docs/upgrade-to-v2.md)

timayabi2020:
﻿

Fixes #1559


tools/Migration/ExamplesMigration.ps1
```diff
@@ -16,14 +16,17 @@ function Start-Copy {
         [ValidateNotNullOrEmpty()]
         [string] $Module = "Users"
     )
+    
+    ../GenerateModules.ps1 -ModuleToGenerate "Teams"
```
peombwa, line 20: @timayabi2020, let's move the examples for the remaining modules in bulk, now that we are confident with the examples migration script. Moving the examples one module at a time will take a while.

The goal should be to have all examples migrated in a single v2.x preview release so we can close https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1559.

peombwa (approved):

PR #1774, opened 2023-01-30
Updated script to log missing external dos url and added optional nam…

timayabi2020:
﻿

Fixes #
Automated update on example files (scrapping examples from api reference docs) is not working for all documented commands because of missing external docs url parameter from open api files. the automation depends on this parameter in order to get examples. We need to have a way of logging such commands for specific troubleshooting.

The figures below show uri paths that have and doesn't have the external docs url parameter respectively.

**Has an external docs url parameter**
image

**Doesn't have an external docs url parameter**
image

- Update the example generator script to allow logging of commands that miss the external docs url parameter 
- Provide a manual way of executing the script provided that the external docs/api reference url is known
- Append "microsoft.graph" namespace to retry fetching the external docs url for commands that missed the parameter on the first fetch, because there is no way of knowing if a command is an action or function.

### DIFF
--- a/tools/ExamplesGenerator.ps1
+++ b/tools/ExamplesGenerator.ps1
@@ -6,17 +6,36 @@ Param(
 )
 function Start-Generator {
     Param(
-        $ModulesToGenerate = @()
+        $ModulesToGenerate = @(),
+        [ValidateNotNullOrEmpty()]
+        [string] $GenerationMode = "auto",
+        [string] $GraphCommand = "GetMg-User",
+        [string] $GraphModule = "Users",
+        [string] $ProfilePath = "v1.0",
+        [string] $ManualExternalDocsUrl = "https://learn.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0&tabs=powershell"
     )
 
     $GraphMapping = @{
         "v1.0" = "examples\v1.0"
         "beta" = "examples\v1.0-beta"
     }
-    $GraphMapping.Keys | ForEach-Object {
-        $graphProfile = $_
-        Get-FilesByProfile -GraphProfile $graphProfile -GraphProfilePath $GraphMapping[$graphProfile] -ModulesToGenerate $ModulesToGenerate 
+    if ($GenerationMode -eq "auto") {
+        $GraphMapping.Keys | ForEach-Object {
+            $graphProfile = $_
+            Get-FilesByProfile -GraphProfile $graphProfile -GraphProfilePath $GraphMapping[$graphProfile] -ModulesToGenerate $ModulesToGenerate 
+        }
     }
+    else {
+          
+        $ProfilePathMapping = "examples\v1.0"
+        if($ProfilePath -eq "beta"){
+            $ProfilePathMapping = "examples\v1.0-beta"
+        }
+        $ModulePath = Join-Path $PSScriptRoot "..\src\$GraphModule\$GraphModule\$ProfilePathMapping"
+        Get-ExternalDocs-Url -ManualExternalDocsUrl $ManualExternalDocsUrl -GenerationMode $GenerationMode -GraphProfilePath $ModulePath -Command $GraphCommand -GraphProfile $ProfilePath -Module -$Module
+            
+    }
+
 }
 function Get-FilesByProfile {
     Param(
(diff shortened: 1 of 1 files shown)


tools/ExamplesGenerator.ps1
```diff
@@ -6,17 +6,36 @@ Param(
 )
 function Start-Generator {
     Param(
-        $ModulesToGenerate = @()
+        $ModulesToGenerate = @(),
+        [ValidateNotNullOrEmpty()]
+        [string] $GenerationMode = "auto",
+        [string] $GraphCommand = "GetMg-User",
```
peombwa, line 12:
```suggestion
        [string] $GraphCommand = "Get-MgUser",
```

tools/ExamplesGenerator.ps1
```diff
@@ -6,17 +6,36 @@ Param(
...
+        }
+    }
+    else {
+          
+        $ProfilePathMapping = "examples\v1.0"
+        if($ProfilePath -eq "beta"){
+            $ProfilePathMapping = "examples\v1.0-beta"
+        }
+        $ModulePath = Join-Path $PSScriptRoot "..\src\$GraphModule\$GraphModule\$ProfilePathMapping"
+        Get-ExternalDocs-Url -ManualExternalDocsUrl $ManualExternalDocsUrl -GenerationMode $GenerationMode -GraphProfilePath $ModulePath -Command $GraphCommand -GraphProfile $ProfilePath -Module -$Module
```
peombwa, line 35: Please rename `Get-ExternalDocs-Url` function to be compliant with PowerShell naming convention.
```suggestion
        Get-ExternalDocsUrl -ManualExternalDocsUrl $ManualExternalDocsUrl -GenerationMode $GenerationMode -GraphProfilePath $ModulePath -Command $GraphCommand -GraphProfile $ProfilePath -Module -$Module
```

tools/ExamplesGenerator.ps1
```diff
@@ -29,15 +48,15 @@ function Get-FilesByProfile {
     )
 
 
-    $ModulesToGenerate | ForEach-Object {
-        $ModuleName = $_
+     $ModulesToGenerate | ForEach-Object {
+         $ModuleName = $_
```
peombwa, line 52: Check formatting for this block.

tools/ExamplesGenerator.ps1
```diff
@@ -98,24 +121,84 @@ function Get-ExternalDocs-Url {
...
+        [string] $Module = "Users",
+        [string] $GenerationMode = "auto",
+        [string] $ManualExternalDocsUrl,
         [ValidateNotNullOrEmpty()]
         [string] $Command = "Get-MgUser",
         [Hashtable] $OpenApiContent,
+        [System.Object] $Method = "GET",
         [string] $GraphProfilePath = (Join-Path $PSScriptRoot "..\src\Users\Users\examples\v1.0")
     )
+    $MissingExternalDocsUrlFolder = Join-Path $PSScriptRoot "..\openApiDocs\MissingExternalDocsUrl\$Module.csv"
```
peombwa, line 133: @timayabi2020, who is this file meant for? Should it be in the docs or metadata repo instead? External docs link is used by all SDKs and GE, not just PowerShell. Limiting this file to just PowerShell may not be ideal.

timayabi2020: @peombwa the report can be shared by just directing someone to this repository upon request.  Since we are the ones maintaining the reports, which happens after every weekly example generation pipeline run., I don't see how we are limiting the report to just PowerShell. Also, the pipeline runs after the weekly open Api docs refresh pipeline and this ensures that the reports are up to date.

peombwa: I see. We should then put the report in https://github.com/microsoftgraph/msgraph-metadata/tree/master/openapi. The sliced OpenAPI documents in this repo are only used by PowerShell. Missing external docs is beyond the scope of this repo as it originates from the metadata and cuts across all SDKs.

To centralize where we track issues, I'd say let's move the report generation to the metadata repo, where it is actionable. Ideally, we should run the report for the entire MS Graph API surface area, which only exists in https://github.com/microsoftgraph/msgraph-metadata/tree/master/openapi. This can be an independent AzDo pipeline in the metadata repo that runs after metadata to OpenAPI conversion.

timayabi2020: As discussed, I have removed the responsibility of generating the missing external docs report from that script.

peombwa (approved):

PR #1836, opened 2023-02-23
Enhancements/v1 tov2 migration

### LINKED ISSUE #1796: Provide v1.x to v2.x migration advisor script
The migration script should provide a find and replace experience for renaming `v1.x` beta commands to `v2.x` beta commands as stated in the [v2 migration guide](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/features/2.0/docs/upgrade-to-v2.md#breaking-changes).

timayabi2020:
﻿

Fixes #1796 

- Addition of a migration module that scans through scripts and outputs suggestions that guides the user on where to change in order to conform to the naming convention of commands i.e. Beta commands
- The module can also modify the script and save the update to a path as defined i.e., Path value passed to ``-UpdatedFilePath`` parameter
- Sample outcome of using the module is as below

image


src/Tools.Migration/test/Microsoft.Graph.Tools.Migration.Tests.ps1
```diff
@@ -0,0 +1,58 @@
+using module "..\Microsoft.Graph.Tools.Migration.psm1"
+Describe "New-MgMigrationPlan Command" {
```
peombwa, line 2: Let's also add test cases to cover scenarios where `# Select-MgProfile beta`/ commands are commented out.

src/Tools.Migration/Microsoft.Graph.Tools.Migration.psm1
```diff
@@ -0,0 +1,225 @@
...
+    # TODO: Settle on object members.
+
+    # Proposed Output
+    # Order | Location (file name:line:tab) | Type (Cmdlet|CmdletParameter|ModuleName) | ApiVersion | Original | New
+ 
+    $i = 0
+    foreach ($_ in $ScriptContent) {
+        $LineContent.Add($_)
+
+        if ($_ -ieq "Select-MgProfile beta" -or $_ -ieq "Select-MgProfile 'beta'" -or $_ -ieq 'Select-MgProfile "beta"' -or $_ -ieq "Select-MgProfile -Name beta" -or $_ -ieq "Select-MgProfile -Name 'beta'" -or $_ -ieq 'Select-MgProfile -Name "beta"') {
```
peombwa, line 115: Can we use REGEx for this match?

src/Tools.Migration/Microsoft.Graph.Tools.Migration.psm1
```diff
@@ -0,0 +1,225 @@
...
+    # TODO: Store changes in an object.
+    # TODO: Write object as output.
+    # TODO: Settle on object members.
+
+    # Proposed Output
+    # Order | Location (file name:line:tab) | Type (Cmdlet|CmdletParameter|ModuleName) | ApiVersion | Original | New
+ 
+    $i = 0
+    foreach ($_ in $ScriptContent) {
+        $LineContent.Add($_)
```
peombwa, line 113: Let's skip the line if it is commented out. i.e., starts with `#`. The logic should trim whitespaces.

src/Tools.Migration/Microsoft.Graph.Tools.Migration.psd1
```diff
@@ -0,0 +1,134 @@
...
+#
+# Generated by: Microsoft
+#
+# Generated on: 2/10/2023
+#
+
+@{
+
+# Script module or binary module file associated with this manifest.
+RootModule = '.\Microsoft.Graph.Tools.Migration.psm1'
```
peombwa, line 12: Should we have `Microsoft.Graph.Tools.Migration` module in a separate GitHub repository? It is currently detached from PowerShell SDK and can be used without a taking a dependency on the SDK. 

cc\ @maisarissi, @ddyett, @timayabi2020

timayabi2020: > Should we have `Microsoft.Graph.Tools.Migration` module in a separate GitHub repository? It is currently detached from PowerShell SDK and can be used without a taking a dependency on the SDK.
> 
> cc\ @maisarissi, @ddyett, @timayabi2020

I agree. We should have the tool in a separate GitHub repository

maisarissi: Sure. If it's detached from the SDK and can be used without taking a dependency, we can have it in a different repository. Also, I believe this is beneficial to the size of this PowerShell module, in case one doesn't need/wanna use the advisor.

timayabi2020:
Closing this PR as it has been moved to this [repo](https://github.com/microsoftgraph/msgraph-tools-migration)

PR #1841, opened 2023-02-28
Update New-MgTeamTagMember.md

YuriySamorodov:
An example for `New-MgTeamTagMember` was incorrect. Made it a bit more clear.


timayabi2020:
@YuriySamorodov thanks for improving this example. Its indeed clearer than what we currently have.

timayabi2020 (approved):

PR #1891, opened 2023-03-21
Pagination support for delta commands

### LINKED ISSUE #1876: Paging support fordelta commands
There is need to add paging support for delta commands.  This is because the response body for delta requests that returns more than one results page contains "@odata.nextLink".

image

timayabi2020:
﻿

Fixes #1876


src/readme.graph.md
```diff
@@ -579,13 +579,13 @@ directive:
   - from: source-file-csharp
     where: $
     transform: >
-      if (!$documentPath.match(/generated%2Fcmdlets%2FGet\w*_List\d*.cs/gm))
+      if (!$documentPath.match(/generated%2Fcmdlets%2FGet\w*_List\d*.cs/gm) || !$documentPath.match(/generated%2Fcmdlets%2FGet\w*_delta\d*.cs/gm))
```
peombwa, line 582: Let's simplify the REGEx to:
```suggestion
      if (!$documentPath.match(/generated%2Fcmdlets%2FGet\w*_(List|Delta)\d*.cs/gm))
```

peombwa (approved):
The change looks good! Please do the same for v2 preview.

Issue #1922, opened 2023-04-05
Connect-MGGraph : You specified a different tenant - once in WithAuthority() and once using WithTenant().


NikCharlebois:
We are getting a weird behavior when trying to connect via Connect-MgGraph, right after connecting to PnP.PowerShell.

Running the following:
```PowerShell
Connect-PnPOnline -ClientId $Global:O365DSCAppId -Tenant $Global:O365DSCTenantId -Thumbprint $Global:O365DSCThumbprint `
    -Url 'https://O365DSC-admin.sharepoint.com' -AzureEnvironment Production
Connect-MGGraph -ClientId $Global:O365DSCAppId -TenantId $Global:O365DSCTenantId -CertificateThumbprint $Global:O365DSCThumbprint
```
Will always result in the following error being thrown:
![image](https://user-images.githubusercontent.com/2547149/230075725-18a7a9ab-93c4-4370-90a9-1d18ea1ffc54.png)

peombwa:
@NikCharlebois, I'm not able to reproduce the issue using the latest versions of both modules. The latest version of PowerShell PnP will only run on PowerShell 7+, which is what I've used to test both modules.

#### Using MS Graph PowerShell v1.25.0 and PnP.PowerShell v2.1.1
![image](https://user-images.githubusercontent.com/7061532/231600483-85577790-6867-4f1c-acfe-4fc666af269d.png)

#### Usings MS Graph PowerShell v2 Preview8 and PnP.PowerShell v2.1.1
![image](https://user-images.githubusercontent.com/7061532/231600342-9ca967c6-6d8c-4a35-a88f-27cc81a776c7.png)

erbAbtis:
I have the same issue in my script which uses the `Microsoft365DSC` module: I want to create a service principal with a certificate thumbprint and then extract the application registration info. After that, I use the `Export-M365DSCConfiguration` function which generates an exception in `MicrosoftGraph.psm` line 76 inside the `MSCloudLoginAssistant` module.

erbAbtis:
> Im not sure about the fix, but for a workaround, does it work if you use separate Runspaces for each one? So you would have a runspace connected to PnPOnline, and another connected to MgGraph.

The workaround with runspace did not work for me. Is there something particular I have to set in order to ensure isolation?

KG-droid:
[...]

PR #2073, opened 2023-06-07
Use authorityUrl instead of tenantId when initializing MSAL

peombwa:
﻿

Fixes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1922

- Use `WithAuthority(authorityUrl)` when initializing MSAL applications to avoid the exception at https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1922. Some versions of MSAL will throw an exception in prod environments when `WithAuthority` and `WithTenantId` are specified on the same client. `GetAuthorityUrl(authContext);` will construct an authority URL that takes into account the specified tenant id.

See https://learn.microsoft.com/en-us/azure/active-directory/develop/msal-client-application-configuration#how-to-specify-the-audience-in-your-codeconfiguration for more info.


andrueastman (approved):
👍🏼

PR #2092, opened 2023-06-15
Update HttpRequest Clone Method

### LINKED ISSUE #2064: New-MgGroup - Randomly occurring Exception during group creation - Microsoft.Graph v. 2.0.0-rc3, Powershell 5.1
Reproduction steps:

``` powershell
$requiredModules = @(
  'Microsoft.Graph.Authentication'
  'Microsoft.Graph.Groups'
)

$jobs = @()

1..50 | ForEach-Object {
$jobs += Start-RSJob -Name "Add group: {$_}" -ModulesToImport $requiredModules -Throttle 10 -ScriptBlock {
    $grp = New-MgGroup -DisplayName "new-mggroup-thread-test" -MailEnabled:$false -MailNickname 'unused' -SecurityEnabled
    if ($null -eq $grp)
    {
      return "Null received from New-MgGroup!" 
    } 
    else
    {
      return "Group {$_} added successfully."
    }
  }
}

while ($jobs.Completed -contains $False) {
    Start-Sleep -Seconds 1
}

$jobs | Wait-RSJob | Receive-RSJob
```
Result:

Sometimes when creating groups in parallel, a few requests fail with errors:
``` powershell
> $jobs | Wait-RSJob | Receive-RSJob

One or more errors occurred.
Null received from New-MgGroup!
Group 2 added successfully.
Group 3 added successfully.
Group 4 added successfully.

> $jobs
Id       Name                 State           HasMoreData  HasErrors    Command
--       ----                 -----           -----------  ---------    -------
117      Add group: {1}       Completed       True         True         ...
118      Add group: {2}       Completed       True         False        ...
119      Add group: {3}       Completed       True         False        ...
120      Add group: {4}       Completed       True         False        ...
```

The exception message retrieved from the job object:
``` powershell
Message        : Collection was modified; enumeration operation may not execute.
Data           : {}
InnerException :
TargetSite     : Void ThrowInvalidOperationException(System.ExceptionResource)
StackTrace     :    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
                    at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
                    at System.Net.Http.Headers.HttpHeaders.<GetEnumerator>d__20.MoveNext()
                    at Microsoft.Graph.PowerShell.HttpMessageLogFormatter.<CloneAsync>d__0.MoveNext()
                 --- End of stack trace from previous location where exception was thrown ---
                    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
                    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
                    at Microsoft.Graph.PowerShell.HttpMessageLogFormatter.<GetHttpRequestLogAsync>d__1.MoveNext()
HelpLink       :
Source         : mscorlib
HResult        : -2146233079
```

----------
The errors occur randomly - I once ran 100 parallel requests of which all succeed, and the other time I had 2 errors in 50 requests.

Should the Powershell Graph be considered as thread safe, or is multithreading not officially supported?

---------

Update:

I was able to reproduce the same error using the built-in `Start-Job | Wait-Job | Receive-Job` commands.

I am however __unable__ to reproduce this issue on Powershell 7.3, using neither `Start-Job`, nor `ForEach-Object -Parallel`.

--------

Update 2:

While I initially reported that the issue is caused by parallel execution, I recently discovered that it randomly occurs during synchronous command calls as well.

The following code has a high probability of producing the exception mentioned above:
``` powershell
1..100 | ForEach-Object {
  New-MgGroup -DisplayName "new-mggroup-test" -MailEnabled:$false -MailNickname 'unused'
}
```

peombwa:
﻿

Fixes #2064 by addressing thread safety intermittent bug in .NET framework when copying request content header.

- Updates how request content headers are copied when cloning a request.

### DIFF
--- a/src/Groups/v1.0/test/New-MgGroup.Recording.json
+++ b/src/Groups/v1.0/test/New-MgGroup.Recording.json
@@ -1,0 +1,35 @@
+{
+  "New-MgGroup+[NoContext]+ShouldCreateNewGroup+$POST+https://graph.microsoft.com/v1.0/groups+1": {
+    "Request": {
+      "Method": "POST",
+      "RequestUri": "https://graph.microsoft.com/v1.0/groups",
+      "Content": "{\r\n  \"displayName\": \"new-mggroup-test\",\r\n  \"mailEnabled\": false,\r\n  \"mailNickname\": \"unused\",\r\n  \"securityEnabled\": true\r\n}",
+      "isContentBase64": false,
+      "Headers": {
+      },
+      "ContentHeaders": {
+        "Content-Type": [ "application/json" ],
+        "Content-Length": [ "123" ]
+      }
+    },
+    "Response": {
+      "StatusCode": 200,
+      "Headers": {
+        "Transfer-Encoding": [ "chunked" ],
+        "Vary": [ "Accept-Encoding" ],
+        "Strict-Transport-Security": [ "max-age=31536000" ],
+        "request-id": [ "86476695-f973-44df-8964-2a53227404ab" ],
+        "client-request-id": [ "aeceaf2b-98af-4fc0-a276-577a2e182727" ],
+        "x-ms-ags-diagnostic": [ "{\"ServerInfo\":{\"DataCenter\":\"West US 2\",\"Slice\":\"E\",\"Ring\":\"1\",\"ScaleUnit\":\"001\",\"RoleInstance\":\"MW2PEPF0000749D\"}}" ],
+        "WWW-Authenticate": [ "Bearer realm=\"\", authorization_uri=\"https://login.microsoftonline.com/common/oauth2/authorize\", client_id=\"00000003-0000-0000-c000-000000000000\"" ],
+        "Date": [ "Mon, 19 Jun 2023 23:03:34 GMT" ]
+      },
+      "ContentHeaders": {
+        "Content-Type": [ "application/json" ],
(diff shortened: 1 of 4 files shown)


jberezanski-mdg:
I'm not sure this will fix the problem. The change from
```
 foreach (var contentHeader in originalRequest.Content.Headers)
```
to
```
originalRequest.Content.Headers?.ToList()
```
still iterates the Dictionary contained inside HttpHeaders. It might execute slightly faster, so the time window in which some other thread could cause the headers dictionary to be modified is narrower and the chance of encountering the InvalidOperationException is smaller, but I believe it is still nonzero.

Looking at the decompiled sources of HttpHeaders and Dictionary in .NET 4.8.1, the only reliable (but ugly) workaround I can see right now is to wrap `originalRequest.Content.Headers?.ToList()` in a `try...catch(InvalidOperationException)` and retry if the exception happens.

peombwa:
Thanks @jberezanski-mdg for the suggestion. That's right! Adding a retry once on `InvalidOperationException` appears to be the only viable workaround for .NET Framework. This shouldn't be an issue for customers running on PowerShell 7.x (.NET 7).

timayabi2020 (approved):

PR #2099, opened 2023-06-20
Fixed non-working, autogenerated example and added a new one

merill:
﻿

Fixes #

- Fixed non-working, autogenerated example and added a new one

### Other links

-
-
-

### DIFF
--- a/src/Identity.Governance/Identity.Governance/examples/v1.0-beta/New-MgEntitlementManagementConnectedOrganization.md
+++ b/src/Identity.Governance/Identity.Governance/examples/v1.0-beta/New-MgEntitlementManagementConnectedOrganization.md
@@ -1,18 +1,57 @@
-### Example 1: Add a connected organization using the DomainName parameter
+### Example 1: Using the New-MgEntitlementManagementConnectedOrganization Cmdlet
+
 ```powershell
-New-MgBetaEntitlementManagementConnectedOrganization -DomainName microsoft.com
+$params = @{
+    displayName = "Connected organization name"
+    description = "Connected organization description"
+    identitySources = @(
+        @{
+            "@odata.type" = "#microsoft.graph.domainIdentitySource"
+            domainName = "example.com"
+            displayName = "example.com"
+        }
+    )
+    state = "proposed"
+}
+
+New-MgEntitlementManagementConnectedOrganization -BodyParameter $params
 ```
 
-### Example 2: Add a connected organization using the IdentitySources parameter
+This example shows how to use the New-MgEntitlementManagementConnectedOrganization Cmdlet with the BodyParameter.
+
+### Example 2: Add a connected organization using the DomainName parameter in the 'Proposed' state
+
+```powershell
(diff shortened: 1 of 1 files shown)


timayabi2020:
@merill thanks for noticing this issue.  Here are my suggestions for the example to work as expected.
1.  Roll back to the previous example 1 title.
2.  Since you are targeting to merge into "dev" branch which is responsible for [V1 releases](https://www.powershellgallery.com/packages/Microsoft.Graph/1.28.0), we don't need "-MgBeta" prefix on cmdlets.  Cmdlets with that prefix are only available on [V2 releases](https://www.powershellgallery.com/packages/Microsoft.Graph/2.0.0-rc3) and the branch responsible is "features/2.0".So remove this ``New-MgBetaEntitlementManagementConnectedOrganization -DomainName microsoft.com``.
3.  Update $params with below.
```
$params = @{
	displayName = "Connected organization name"
	description = "Connected organization description"
	identitySources = @(
		@{
			"@odata.type" = "#microsoft.graph.domainIdentitySource"
			domainName = "example.com"
			displayName = "example.com"
		}
	)
	state = "proposed"
}
```
Check Api reference [here](https://learn.microsoft.com/en-us/graph/api/entitlementmanagement-post-connectedorganizations?view=graph-rest-1.0).
4. Delete the second example
5. I have confirmed that the above request body works
image

timayabi2020 (changes requested):
@merill thanks for noticing this issue.  Here are my suggestions for the example to work as expected.
1.  Roll back to the previous example 1 title.
2.  Since you are targeting to merge into "dev" branch which is responsible for [V1 releases](https://www.powershellgallery.com/packages/Microsoft.Graph/1.28.0), we don't need "-MgBeta" prefix on cmdlets.  Cmdlets with that prefix are only available on [V2 releases](https://www.powershellgallery.com/packages/Microsoft.Graph/2.0.0-rc3) and the branch responsible is "features/2.0".So remove this ``New-MgBetaEntitlementManagementConnectedOrganization -DomainName microsoft.com``.
3.  Update $params with below.
```
$params = @{
	displayName = "Connected organization name"
	description = "Connected organization description"
	identitySources = @(
		@{
			"@odata.type" = "#microsoft.graph.domainIdentitySource"
			domainName = "example.com"
			displayName = "example.com"
		}
	)
	state = "proposed"
}
```
Check Api reference [here](https://learn.microsoft.com/en-us/graph/api/entitlementmanagement-post-connectedorganizations?view=graph-rest-1.0).
4. Delete the second example
5. I have confirmed that the above request body works
image

merill:
@timayabi2020 Using body parameters is not really the best way to do things in PowerShell. Also the other parameters like the name are optional. Do we really need to have them?

Also, is there a reason why we want to remove the second example? I can fix the name of the cmdlet to work with Graph PS v1 beta

merill:
I have put back the body parameter example and also fleshed out the examples. Cheers.

timayabi2020 (approved):

PR #2125, opened 2023-07-06
[v2] Examples Update

timayabi2020:
This pull request was automatically created by Azure Pipelines. **Important** Check for unexpected deletions or changes in this PR.


src/CloudCommunications/beta/examples/Get-MgBetaUserOnlineMeeting.md
```diff
@@ -1,46 +1,34 @@
...
 
-### Example 3: Retrieve an online meeting by joinMeetingId
+### Example 2: Code snippet
 
 ```powershell
 Import-Module Microsoft.Graph.Beta.CloudCommunications
 
 # A UPN can also be used as -UserId.
-Get-MgBetaUserOnlineMeeting -UserId $userId -Filter "joinMeetingIdSettings/joinMeetingId eq '1234567890'"
+Get-MgBetaUserOnlineMeetingAttendeeReport -UserId $userId -OnlineMeetingId $onlineMeetingId
```
peombwa, line 19: The example shown does not match the file name. `Get-MgBetaUserOnlineMeeting.md` vs `Get-MgBetaUserOnlineMeetingAttendeeReport`.

peombwa (changes requested):

src/CloudCommunications/beta/examples/Get-MgBetaUserOnlineMeeting.md
timayabi2020: Thanks for flagging this. I've noticed that the script was checking against the api reference for examples with patterns that match the command. Ideally even if there is a pattern that matches the command, it ought to have checked for an exact match. i.e. Matching a string (Command) within boundaries. 
It also needs to retain correct examples if the script contained a mixture of wrong and correct examples.
The latest commit I've done, addresses the above issues.

tools/ImportExamples.ps1
```diff
@@ -313,7 +313,7 @@ function Update-ExampleFile {
         Clear-Content $ExampleFile -Force
         for ($d = 0; $d -lt $HeaderList.Count; $d++) { 
             $CodeValue = $ExampleList[$d].Trim()
-            if($CodeValue.Contains($CommandPattern)){
+            if($CodeValue-match "\b$CommandPattern\b"){
```
peombwa, line 316: Fix spacing for code readability:
```suggestion
            if ($CodeValue -match "\b$CommandPattern\b") {
```

tools/ImportExamples.ps1
```diff
@@ -358,15 +358,24 @@ function Update-ExampleFile {
...
         }else{
-            if(-not($Content | Select-String -pattern $CommandPattern)){
             Clear-Content $ExampleFile -Force
+            if($Content | Select-String -pattern $CommandPattern){
+                Retain-ExistingCorrectExamples -Content $Content -File $ExampleFile
+            }else{
             #Replace everything with boiler plate code
-            $DefaultBoilerPlate = "### Example 1: {{ Add title here }}`r`n``````powershell`r`n PS C:\> {{ Add code here }}`r`n`n{{ Add output here }}`r`n```````n`n{{ Add description here }}`r`n`n### Example 2: {{ Add title here }}`r`n``````powershell`r`n PS C:\> {{ Add code here }}`r`n`n{{ Add output here }}`r`n```````n`n{{ Add description here }}`r`n`n"
-            Add-Content -Path $ExampleFile -Value $DefaultBoilerPlate.Trim()
+                $DefaultBoilerPlate = "### Example 1: {{ Add title here }}`r`n``````powershell`r`n PS C:\> {{ Add code here }}`r`n`n{{ Add output here }}`r`n```````n`n{{ Add description here }}`r`n`n### Example 2: {{ Add title here }}`r`n``````powershell`r`n PS C:\> {{ Add code here }}`r`n`n{{ Add output here }}`r`n```````n`n{{ Add description here }}`r`n`n"
```
peombwa, line 366: It looks like this logic of removing example placeholders isn't working as per this change in your PR - https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/80dcb83c163b428f266fe8988d0851771fc90b38/src/Calendar/beta/examples/New-MgBetaUserEventExtension.md.

You can for example consider a generic implementation like:
- Trim whitespaces from both ends of the string content (md file with placeholder text).
- Find and replace instances where the string content starts with `Example .*: {{ Add title here }}` and ends in `{{ Add description here }}`.

timayabi2020: The objective of that particular block is not to remove the boiler plate code. It instead sets it as a content to the md file incase the file contains nothing that matches the command name and also if whatever the content we are getting from the api reference does not match the command name.  It therefore corrects any historical errors in regards to files that have examples not related to the command names.

peombwa: @timayabi2020, what's the effect of leaving the example blank instead? We shouldn't be showing boiler plate text when an example doesn't exist in the cmdlet reference docs. Are removing the boiler plate text in the SDK docs repro by any chance? If not, please create an issue to track this.

peombwa (approved):

tools/ImportExamples.ps1
timayabi2020: Yes I am removing them. However, I will create another issue to transfer that functionality from the reference docs pipeline to this example import script

PR #2183, opened 2023-07-25
Include "List" variant when updating example files

timayabi2020:
Currently the examples import script only fetches and process examples for cmdlets that fall under the "Get" variant. 
This PR will include cmdlets that also fall under the "List" variant.
image

### DIFF
--- a/tools/ImportExamples.ps1
+++ b/tools/ImportExamples.ps1
@@ -3,10 +3,6 @@
 Param(
     $ModulesToGenerate = @(),
     $Available = @(),
-    [hashtable]$V1CommandGetVariantList = @{},
-    [hashtable]$BetaCommandGetVariantList = @{},
-    [hashtable]$V1CommandListVariantList = @{},
-    [hashtable]$BetaCommandListVariantList = @{},
     [string] $ModuleMappingConfigPath = (Join-Path $PSScriptRoot "..\config\ModulesMapping.jsonc"),
     [string] $FolderForExamplesToBeReviewed = (Join-Path $PSScriptRoot "..\examplesreport"),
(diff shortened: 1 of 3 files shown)


tools/ImportExamples.ps1
```diff
@@ -90,10 +90,16 @@ function Get-Files {
...
                     $CommandValue = $BetaCommandGetVariantList[$Command]
+                    if(-not($CommandValue)){
+                        $CommandValue = $BetaCommandListVariantList[$Command]
+                    }
                         
                 }
                 else {
                     $CommandValue = $V1CommandGetVariantList[$Command]
+                    if(-not($CommandValue)){
+                        $CommandValue = $V1CommandListVariantList[$Command]
```
peombwa, line 101: What about other variants? Commands can have more than just `Get` and `List` variants:
``` powershell
➜ Find-MgGraphCommand -Command ".*-MgUser" | select Command, Method, URI, Variants

Command       Method URI              Variants
-------       ------ ---              --------
Get-MgUser    GET    /users/{user-id} {Get, GetViaIdentity}
Get-MgUser    GET    /users           {List}
New-MgUser    POST   /users           {Create, CreateExpanded}
Remove-MgUser DELETE /users/{user-id} {Delete, DeleteViaIdentity}
Update-MgUser PATCH  /users/{user-id} {Update, UpdateExpanded, UpdateViaIdentity, UpdateViaIdentityExpanded}
```
It is also worth noting that variant names for OData actions and functions are based on the command's verb. For example, `Revoke-MgUserSignInSession` will have a variant name of `Revoke` and `RevokeViaIdentity`.

Would it be better to use the HTTP method name to group the commands instead?

PR #2202, opened 2023-07-28
Adds autorest directive to pluralize "person" noun.

### LINKED ISSUE #2201: Wrong cmdlet generated for "/admin/people/pronouns"
Generated cmdlet has a singularized "people" noun as part of the cmdlet subject. i.e., "person" noun instead of "people" as per what is available on the Api path.
image

timayabi2020:
﻿

Fixes #2201 

- Added directive to pluralize the "people" noun


peombwa:
This is by design. The API returns a resource type named `person` - https://learn.microsoft.com/en-us/graph/api/resources/person?view=graph-rest-beta.

PowerShell requires singularization of subjects/nouns in command names whenever possible - https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.3#use-a-specific-noun-for-a-cmdlet-name-sd01:
> To enhance the user experience, the noun that you choose for a cmdlet name should be singular. For example, use the name Get-Process instead of Get-Processes. It is best to follow this rule for all cmdlet names, even when a cmdlet is likely to act upon more than one item.

`*/people*` APIs have always been singularized to `person` even in v1.x - https://learn.microsoft.com/en-us/graph/api/user-list-people?view=graph-rest-beta&tabs=powershell#request.

timayabi2020:
@peombwa thanks for the clarification

peombwa:
Reopening following an internal discussion with the workload owner. There was an API review exception made to `admin/people` and the condition at https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/2202#issuecomment-1655937732 does not apply to the API path.

src/readme.graph.md
```diff
@@ -165,6 +165,11 @@ directive:
       subject: ^(.*)List(.*)(As.*)$
     set:
       subject: $1$2$3
+  - where:
+      verb: Get|New|Remove|Update
```
peombwa, line 169: We can exclude the verb since we want the rename to apply to commands verbs (HTTP methods + OData actions and functions)
```suggestion
```

peombwa (approved):

PR #2206, opened 2023-08-01
Fix error when running on Azure Functions

### LINKED ISSUE #2198: Microsoft.Graph.Authentication fails to load in Azure Functions with WEBSITE_RUN_AS_PACKAGE
**Describe the bug**
When running The Graph SDK module in Azure functions as a package, the authentication module fails to load, causing all other modules to also fail to load.

> ERROR: Could not find file 'C:\home\site\wwwroot\\.mg'

I have traced the issue to this line of code:
https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/8b8b027dfef5b7ce1b744f4b8258ef4ac93e63d0/src/Authentication/Authentication/Common/GraphSessionInitializer.cs#L33

The issue is that when running `WEBSITE_RUN_AS_PACKAGE`, the user home directory is read-only, so it fails to create the directory.

**Workaround**
You can work around this issue by creating a `.mg` directory in the root of your function project.

**To Reproduce**
1. Create an Azure function using the powershell 7 runtime on windows, use EP1 plan. (not consumption)
1. Create a azure powershell function using the VS Code powershell extension
1. Edit requirements.psd1 to include a few v2 Graph modules, including `Microsoft.Graph.Authentication`
1. Add `Import-Module Microsoft.Graph.Authentication` somewhere in your function
1. Deploy to function app using VS Code
1. Test function, see errors.

**Expected behavior**
Module should load without errors

**Module Version**
2.2.0

**Additional context**
I'm wondering if #2133 is also related to this issue. Azure automation potentially obfuscating this error?

BHoggs:
Fixes #2198

The fix for this turned out to be quite simple - just added a try block around the code that attempts to create the graph directory in the user's profile.

As far as I can see this has no flow-on effects. This folder is not used for anything else, other than checking if an existing config file exists. However this step does not fail with the change I have made.

Have tested in Azure Functions ~4, 7.2 runtime.

I also tested in Azure Automation, but this unfortunately does not fix the import error there.


src/Authentication/Authentication/Common/GraphSessionInitializer.cs
```diff
@@ -30,7 +30,13 @@ internal static GraphSession CreateInstance(IDataStore dataStore = null)
         {
             IGraphOption graphOptions = null;
             // Try to create directory if it doesn't exist.
-            Directory.CreateDirectory(Core.Constants.GraphDirectoryPath);
+            try 
+            {
+                Directory.CreateDirectory(Core.Constants.GraphDirectoryPath);
+            }
+            finally
```
peombwa, line 37: I believe you meant to use catch instead of catch. Finally will still bubble to exception to the caller:
```suggestion
            catch
```

peombwa:
Thanks for your contribution!

Can you confirm if you are able to successfully login with the try catch block added - `Connect-MgGraph ...`? `.mg` directory is used by the auth library to sync "auth records" for token caching. I suspect the auth library will default to process context (in-memory cache) if the directory does not exist.

src/Authentication/Authentication/Common/GraphSessionInitializer.cs
BHoggs: Ah, right you are. I actually tested this with a catch block originally, but swapped it with finally when I didn't think it was needed. Put it back to catch now.

BHoggs:
Welcome! I guess in 99% of scenarios this doesn't change anything. It's pretty rare where the home directory is read-only. As far as I could tell from my testing it appears to work fine on Azure functions in this scenario, and doesn't effect other scenarios.

I did only test the function app scenario using managed identity (`Connect-MgGraph -Identity`), do you think it also needs to be tested using client/secret auth?

peombwa (approved):
LGTM! I've validated that everything works as expected in Azure Function using certificates, client secrets, and managed identities.

PR #2317, opened 2023-09-21
Translates /me to /users/{user-id}

### LINKED ISSUE #2306: 'find-MgGraphCommand -Uri ' not working for /me endpoints
**Describe the bug**
There is a customer [complaint ]( https://github.com/microsoftgraph/microsoft-graph-docs-contrib/issues/50)that they cannot find the permissions for a /me endpoint using find-MgGraphCommand -Uri. On testing, the /me endpoints are currently not working.

**To Reproduce**
1. Run find-MgGraphCommand -Uri '/me/events' or find-MgGraphCommand -Uri '/me/mailboxSettings'
2. Error thrown is
find-MgGraphCommand : URI ' https://graph.microsoft.com/v1.0/me/mailboxSettings' in  is not valid or is not currently supported by
the SDK. Ensure the URI is formatted correctly and try again.
3. Using find-MgGraphCommand -command Get-MgUserMailboxSetting works

**Expected behavior**
expected to return the command, URI and permissions

**Debug Output**
find-MgGraphCommand : URI ' https://graph.microsoft.com/v1.0/me/events' in  is not valid or is not currently supported by the SDK.
Ensure the URI is formatted correctly and try again.
At line:1 char:1
+ find-MgGraphCommand -Uri '/me/events' -Debug
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Find-MgGraphCommand

timayabi2020:
Fixes #2306 

- Replaces /me path segments with /users/{user-id}

image

### DIFF
--- a/src/Authentication/Authentication/custom/Find-MgGraphCommand.ps1
+++ b/src/Authentication/Authentication/custom/Find-MgGraphCommand.ps1
@@ -180,9 +180,16 @@ Function Find-MgGraphCommand {
             if (!$GraphUri.IsAbsoluteUri) {
                 $GraphUri = GraphUri_ConvertRelativeUriToAbsoluteUri -Uri $GraphUri -ApiVersion $ApiVersion
             }
-           
-            $Segment = $GraphUri.Segments[2]
-            if ($Segment.StartsWith("me")) {
+            
+            $ContainsMeSegment = $False
+            $Segment = $GraphUri.Segments
+            foreach ($s in $Segment) {
+                if ($s.StartsWith("me")) {
+                    $ContainsMeSegment = $True
+                    break
+                }
+            }
+            if ( $ContainsMeSegment) {
                 $GraphUri = $GraphUri.AbsoluteUri.Replace("/me/", "/users/{id}/")
             }
             Write-Debug "Resolved URI: $GraphUri."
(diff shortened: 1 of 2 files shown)


src/Authentication/Authentication/custom/Find-MgGraphCommand.ps1
```diff
@@ -187,7 +187,10 @@ Function Find-MgGraphCommand {
                 [Parameter(Mandatory = $true, Position = 0)]
                 [System.Uri]$Uri
             )
-
+            #Check if Uri contains /me and replace it with /users/{user-id}
+            if ($Uri.AbsoluteUri.Contains("/me/")) {
```
calebkiage, line 191: I think it's better to use a relative to make the Uri relative to the graph api root and check `$Uri.StartsWith("/me/)`. That way, you'll not have a problem with urls like `https://graph.microsoft.com/v1.0/profile/me/` if it ever existed in future.

calebkiage, line 191: I think it's better to use a relative to make the Uri relative to the graph api root and check `$Uri.StartsWith("/me/)`. That way, you'll not have a problem with urls like `https://graph.microsoft.com/v1.0/profile/me/` if it ever existed in future.

calebkiage (changes requested):

src/Authentication/Authentication/custom/Find-MgGraphCommand.ps1
timayabi2020: Thanks for noticing this. I have made changes to handle cases where -Uri parameter can be passed as either relative or absolute.

PR #2359, opened 2023-10-12
Remove Update...Photo cmdlets

mackoo13:
The owner team of People workload is performing a cleanup of MS Graph cmdlets. Several cmdlets have never been supported by the underlying API and were never intended to be created by MS Graph automation.

Update...Photo cmdlets return 405. These cmdlets trigger a request to /photo endpoint, which is not allowed by design. Instead, there are other cmdlets to update an image on the correct /photo/$value path, e.g. Set-MgUserPhotoContent.

The change has been validated locally.

Contact in MS: makocot


timayabi2020:
Generation of any MS Graph cmdlet is as a result of availability of a path in the metadata. I suspect this issue occurs in other SDKs too. So, it should be handled at the source (metadata). I would also suggest that you liaise with the API reviewer for any clarifications needed.

peombwa (approved):
The PR looks good to me as temporary fix for PowerShell. Per internal comms., CSDL annotations will need to be added to remove the endpoints upstream when the capability is available.

PR #2361, opened 2023-10-12
Fix escaping of filter values with displayName

### LINKED ISSUE #2280: Get-MGGroup filter with special characters doesn't work
Hi,

we have a group with a "+" and a "|" char and im unable to filter this group.

I tried to encode the searchstring, but nothing helped. 
The "+" sign can be encoded with %2b, but the "|" char is unsearchable.

So i tried this:
```
$str = "P+R | Workspace"
Get-MgGroup -Filter "displayName eq '$([System.Web.HttpUtility]::UrlDecode($str))'"
```

With Invoke-MgGraphRequest everything is working fine.
```
(Invoke-MgGraphRequest -Uri "https://graph.microsoft.com/v1.0/groups?filter=displayname eq '$([System.Web.HttpUtility]::UrlDecode($str))'").value
```

I'm using Powershell Graph v.2.4.0

timayabi2020:
This PR fixes #2280 by preventing filter values with + on the ```displayName``` from being replaced by an empty string.
**Changes** by adding a regex to check for values enclosed within double or single qoutes.
image


tools/Custom/PSCmdletExtensions.cs
```diff
@@ -28,7 +30,7 @@ internal static string UnescapeString(this PSCmdlet cmdlet, string value)
             try
             {
                 var unescapedString = Uri.UnescapeDataString(value);
-                return unescapedString.Replace('+', ' ');
+                return EnclosedStringRegex.IsMatch(value) ? unescapedString: unescapedString.Replace('+', ' ');
```
calebkiage, line 33: I think for this search, a regex pattern isn't really required. You can use `StartsWith` and `EndsWith` here and it should make the intent clearer. It should also perform better.

timayabi2020: For the given solution, an example of a value the extension method is receiving is something like ```displayName eq 'P+R | Workspace'```. So, we are only looking for this part.  ``'P+R | Workspace'`` in that string. The value could assume any pattern. e.g. ``memberOf like 'D+R TestMember'``. For such scenarios, the goal is to ensure that the plus (+) sign inside single quotes are not replaced with an empty space.

timayabi2020: The suggestion has been done and it's working as expected. Thanks

peombwa (approved):

PR #2455, opened 2023-11-21
Escapes special characters on path segments when using ``Invoke-MgGraphRequest``

### LINKED ISSUE #1947: Invoke-MgGraphRequest can't handle URI with #
Hello,

I would like to use the Invoke-MgGraphRequest cmdlet to request Guest users. Guest users' UPN contains '#' characters.
Example: first.last_foo.com#EXT#@contoso.onmicrosoft.com

```
$upn = first.last_foo.com#EXT#@contoso.onmicrosoft.com
Invoke-MgGraphRequest -Method GET -Uri https://graph.microsoft.com/beta/users/$upn
Invoke-MgGraphRequest : GET https://graph.microsoft.com/beta/users/first.last_foo.com
HTTP/1.1 404 Not Found
```

If I request the same URI on MS Graph Explorer, it works well.
It seems that Invoke-MgGraphRequest cut the URI at the first '#' character.

I've tried different variations, but issue persists:
```
Invoke-MgGraphRequest -Method GET -Uri "https://graph.microsoft.com/beta/users/$upn"
```
```
$uri = 'https://graph.microsoft.com/beta/users/first.last_foo.com#EXT#@contoso.onmicrosoft.com'
Invoke-MgGraphRequest -Method GET -Uri $uri
```

Could you please have a check and fix it?

Module version tested: 1.20.0 and 1.25.0.

Best,
julmsy

timayabi2020:
﻿

Fixes #1947

-  Added function to escape data string/url encode path segments that have special characters like #. Results for both Absolute and relative uri are shown on the image below.
image

### DIFF
--- a/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
+++ b/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
@@ -23,7 +23,8 @@ namespace Microsoft.Graph.PowerShell.Authentication.Helpers
                 //Skips the left part of the uri i.e https://graph.microsoft.com
                 if (counter > 2)
                 {
-                    sb.Append("/" + Uri.EscapeDataString(segment));
+                    sb.Append('/');
+                    sb.Append(Uri.EscapeDataString(segment));
                 }
                 counter++;
             }
--- a/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
+++ b/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
@@ -1,0 +1,34 @@
+// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+using System;
+using System.Text;
+
+namespace Microsoft.Graph.PowerShell.Authentication.Helpers
+{
+    /// <summary>
+    /// Escape data string/url encode Uris that have paths containing special characters like #.
+    /// For a path like /beta/users/first.last_foo.com#EXT#@contoso.onmicrosoft.com, the last segment contains special characters that need to be escaped
+    /// </summary>
+    public static class EncodeUriPaths
+    {
+        public static Uri EscapeDataStrings(this Uri uri)
+        {
+            int counter = 0;
+            var pathSegments = uri.OriginalString.Split('/');
+            StringBuilder sb = new StringBuilder();
+            foreach (var segment in pathSegments)
+            {
+                //Skips the left part of the uri i.e https://graph.microsoft.com
+                if (counter > 2)
+                {
+                    sb.Append("/" + Uri.EscapeDataString(segment));
+                }
+                counter++;
+            }
+            Uri escapedUri = new Uri(uri.GetLeftPart(UriPartial.Authority) + sb.ToString());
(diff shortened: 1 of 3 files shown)


peombwa (changes requested):
Thanks for making the fix! Please see my suggestions on the PR.

src/Authentication/Authentication/Helpers/EncodeUriPath.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    /// Escape data string/url encode Uris that have paths containing special characters like #.
+    /// For a path like /beta/users/first.last_foo.com#EXT#@contoso.onmicrosoft.com, the last segment contains special characters that need to be escaped
+    /// </summary>
+    public static class EncodeUriPaths
+    {
+        public static Uri EscapeDataStrings(this Uri uri)
+        {
+            int counter = 0;
+            var pathSegments = uri.OriginalString.Split('/');
```
calebkiage, line 19: Why not use `uri.Segments` instead of `uri.OriginalString.Split('/')`?

src/Authentication/Authentication/Helpers/EncodeUriPath.cs
```diff
@@ -0,0 +1,34 @@
...
+            StringBuilder sb = new StringBuilder();
+            foreach (var segment in pathSegments)
+            {
+                //Skips the left part of the uri i.e https://graph.microsoft.com
+                if (counter > 2)
+                {
+                    sb.Append("/" + Uri.EscapeDataString(segment));
```
calebkiage, line 26:
```suggestion
                    sb.Append('/');
                    sb.Append(Uri.EscapeDataString(segment));
```

src/Authentication/Authentication/Helpers/EncodeUriPath.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    public static class EncodeUriPaths
+    {
+        public static Uri EscapeDataStrings(this Uri uri)
+        {
+            int counter = 0;
+            var pathSegments = uri.OriginalString.Split('/');
+            StringBuilder sb = new StringBuilder();
+            foreach (var segment in pathSegments)
+            {
```
calebkiage, line 22: Why not use `uri.Segments` instead of `uri.OriginalString.Split('/')`?

```suggestion
            StringBuilder sb = new StringBuilder();
            foreach (var segment in uri.Segments)
            {

```

timayabi2020: Using uri.Segments will ommit the part with special characters.  Given a url like this https://graph.microsoft.com/beta/users/first.last_foo.com#EXT#@contoso.onmicrosoft.com., You'll only end up getting '/beta/users/first.last_foo.com' yet we want the whole thing including that omitted part with special characters.

calebkiage: It makes sense. So how do you avoid swallowing anchor components in a URL? Because `#` marks the beginning of an anchor.

peombwa (approved):

src/Authentication/Authentication/Helpers/EncodeUriPath.cs
ddyett: good call @calebkiage, turns out this was a problem.

PR #2512, opened 2024-01-03
Access Token Proof of Possession Capability

FehintolaObafemi:
﻿

Fixes #

- Introducing an Access Token Proof of Possession capability
- Updating the dependencies to the latest version

### Other links

- Link to parent work item: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1795829
- Link to design documentation: https://microsoft-my.sharepoint.com/:w:/p/taofeeko/ERh0KZRTuMJKmeltAe8oTzgBkUoZ7wq3yVZAhhuFQ61g-Q?e=RSwT1m


src/Authentication/Authentication.Core/Utilities/PopClient.cs
```diff
@@ -0,0 +1,88 @@
...
+            //_interactiveBrowserCredential = (InteractiveBrowserCredential)credential;
+            _interactiveBrowserCredential = new InteractiveBrowserCredential(new InteractiveBrowserCredentialBrokerOptions(WindowHandleUtlities.GetConsoleOrTerminalWindow()));
+
+            if (!(credential is ISupportsProofOfPossession))
+            {
+                throw new ArgumentException("The provided TokenCredential does not support proof of possession.", nameof(credential));
+            }
+
+            var pipelineOptions = new HttpPipelineOptions(options);
+            pipelineOptions.PerRetryPolicies.Add(new InteractivePopTokenAuthenticationPolicy(_interactiveBrowserCredential, "https://graph.microsoft.com/.default", () => _authenticationRecord));
```
adamedx: What happens if the server doesn't support PoP, it sounds like we fall back to "regular" token.

FehintolaObafemi: This has been resolved. If the server does not support PoP the request fails.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +477,63 @@ public static Task DeleteAuthRecordAsync()
                 File.Delete(Constants.AuthRecordPath);
             return Task.CompletedTask;
         }
+
+        public static PopTokenRequestContext CreatePopTokenRequestContext(IAuthContext authContext)
+        {
+            // Creating a httpclient that would handle all pop calls
+            Uri popResourceUri = GraphSession.Instance.GraphRequestProofofPossession.Uri ?? new Uri("https://canary.graph.microsoft.com/beta/me"); //PPE (https://graph.microsoft-ppe.com) or Canary (https://canary.graph.microsoft.com) or (https://20.190.132.47/beta/me)
+            HttpClient popHttpClient = new(new HttpClientHandler { ServerCertificateCustomValidationCallback = (_, _, _, _) => true });
```
darrelmiller, line 485: This is vulnerable to resource exhaustion.  We should also never disable server certificate validation.

docs/authentication.md
```diff
@@ -112,6 +112,26 @@ When using `-AccessToken`, we won't have access to the refresh token and the cli
 
 Before using the provided `-AccessToken` to get Microsoft Graph resources, customers should ensure that the access token has the necessary scopes/ permissions needed to access/modify a resource.
 
+### Access Token Proof of Possession (AT PoP)
+
+AT PoP is a security mechanism that binds an access token to a cryptographic key that only the intended recipient has. This prevents unauthorized use of the token by malicious actors. AT PoP enhances data protection, reduces token replay attacks, and enables fine-grained authorization policies.
```
darrelmiller, line 117: Who is the "intended recipient" here? My understanding is that it is the token requestor, not the token receiver.

docs/authentication.md
```diff
@@ -112,6 +112,26 @@ When using `-AccessToken`, we won't have access to the refresh token and the cli
...
+
+Note: AT PoP requires WAM to function.
```
darrelmiller, line 119: Can you expand WAM as most people will not know what it is?

docs/authentication.md
```diff
@@ -112,6 +112,26 @@ When using `-AccessToken`, we won't have access to the refresh token and the cli
...
+
+Microsoft Graph PowerShell module supports AT PoP in the following scenario:
+
+- To enable AT PoP on supported devices
+
+```PowerShell
+Set-MgGraphOption -EnableATPoP $true
```
darrelmiller, line 126: Why would we create a whole new cmdlet for this instead of making it a parameter of Connect-MgGraph like all other auth options?

docs/authentication.md
FehintolaObafemi: Done

docs/authentication.md
FehintolaObafemi: Yes, it is indeed the token requestor.

docs/authentication.md
FehintolaObafemi: - We want each user to be able to customize thier msgraph ps without having to run "Connect-MgGraph -EnableATPoP" or "Connect-MgGraph -WAM" everytime the user decides to either use WAM or the AT PoP feature
- We decided to store the user's settings per se in a json file named "MgGraphOption" which to my knowledge would include other setting option moving forward. This was initially done when implementing the WAM feature.

src/Authentication/Authentication.Core/Interfaces/IGraphOptions.cs
```diff
@@ -2,14 +2,11 @@
...
-using System;
-using System.Security;
-using System.Security.Cryptography.X509Certificates;
-
 namespace Microsoft.Graph.PowerShell.Authentication
 {
     public interface IGraphOption
     {
         bool EnableWAMForMSGraph { get; set; }
+        bool EnableATPoPForMSGraph { get; set; }
```
bgavrilMS, line 10: Strictly speaking, adding a property to an interface is a breaking change. Data objects like this should not be backed by interfaces.

src/Authentication/Authentication.Core/Interfaces/IGraphRequestProofofPossession.cs
```diff
@@ -0,0 +1,22 @@
...
+// ------------------------------------------------------------------------------
+
+using Azure.Core;
+using Azure.Identity;
+using System;
+using System.Net.Http;
+
+namespace Microsoft.Graph.PowerShell.Authentication
+{
+    public interface IGraphRequestProofofPossession
```
bgavrilMS, line 12: Prefer to use objects instead of interfaces. You will not be able to add members here.

Also there is no reason why this is an interface. A data object with public setters is perfectly testable.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -113,23 +121,46 @@ private static async Task<InteractiveBrowserCredential> GetInteractiveBrowserCre
...
+                GraphSession.Instance.GraphRequestProofofPossession.PopTokenContext = CreatePopTokenRequestContext(authContext);
+                GraphSession.Instance.GraphRequestProofofPossession.BrowserCredential = interactiveBrowserCredential;
+            }
+
             if (!File.Exists(Constants.AuthRecordPath))
             {
                 AuthenticationRecord authRecord;
-                var interactiveBrowserCredential = new InteractiveBrowserCredential(interactiveOptions);
+                //var interactiveBrowserCredential = new InteractiveBrowserCredential(interactiveOptions);
                 if (IsWamSupported())
```
bgavrilMS, line 143: Your implemenation of `IsWamSupported` is not really correct. WAM is not supported on Win Server 2016. MSAL does have a few helpers like `IsBrokerSupported` and `IsPOPSuppoportedByClient` btw.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -113,23 +121,46 @@ private static async Task<InteractiveBrowserCredential> GetInteractiveBrowserCre
         {
             if (authContext is null)
                 throw new AuthenticationException(ErrorConstants.Message.MissingAuthContext);
-            var interactiveOptions = IsWamSupported() ? new InteractiveBrowserCredentialBrokerOptions(WindowHandleUtlities.GetConsoleOrTerminalWindow()) : new InteractiveBrowserCredentialOptions();
+            var interactiveOptions = IsWamSupported() ?
+                                        new InteractiveBrowserCredentialBrokerOptions(WindowHandleUtlities.GetConsoleOrTerminalWindow()) :
```
bgavrilMS, line 125: If you are an SDK, you should expose the window handle. You can't assume the consuming app is a command line app.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +477,63 @@ public static Task DeleteAuthRecordAsync()
                 File.Delete(Constants.AuthRecordPath);
             return Task.CompletedTask;
         }
+
+        public static PopTokenRequestContext CreatePopTokenRequestContext(IAuthContext authContext)
```
bgavrilMS, line 481: Why public?

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +477,63 @@ public static Task DeleteAuthRecordAsync()
...
+        {
+            // Creating a httpclient that would handle all pop calls
+            Uri popResourceUri = GraphSession.Instance.GraphRequestProofofPossession.Uri ?? new Uri("https://graph.microsoft.com/beta/organization"); //PPE (https://graph.microsoft-ppe.com) or Canary (https://canary.graph.microsoft.com) or (https://20.190.132.47/beta/me)
+            HttpClient popHttpClient = new(new HttpClientHandler());
+
+            // Find the WWW-Authenticate header in the response.
+            var popMethod = GraphSession.Instance.GraphRequestProofofPossession.HttpMethod ?? HttpMethod.Get;
+            var popResponse = popHttpClient.SendAsync(new HttpRequestMessage(popMethod, popResourceUri)).Result;
```
bgavrilMS, line 489: sync over async is really bad for perf and should be avoided.

src/Authentication/Authentication.Core/Interfaces/IGraphOptions.cs
bgavrilMS, line 10: Design question - have you thought to enable this by default and provide an opt-out at most?

docs/authentication.md
bgavrilMS: Both WAM and POP should be the default options in my opinion. And the opt-out of these features obscure.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +477,63 @@ public static Task DeleteAuthRecordAsync()
...
+            var popChallenge = popResponse.Headers.WwwAuthenticate.First(wa => wa.Scheme == "PoP");
+            var nonceStart = popChallenge.Parameter.IndexOf("nonce=\"") + "nonce=\"".Length;
```
bgavrilMS, line 491: Parsing WWW-Authenticate is a bit tricky. MSAL has a parser that is better suited. I think Azure SDK folks wanted to provide one as well

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +477,63 @@ public static Task DeleteAuthRecordAsync()
...
+                case "PATCH":
+                    return RequestMethod.Patch;
+                case "TRACE":
+                    return RequestMethod.Trace;
+                default:
+                    throw new ArgumentException($"Unsupported HTTP method: {httpMethod.Method}");
+            }
+        }
+    }
+    internal class PopClientOptions : ClientOptions
```
bgavrilMS, line 536: ?

src/Authentication/Authentication/Handlers/AuthenticationHandler.cs
```diff
@@ -63,9 +65,24 @@ private async Task AuthenticateRequestAsync(HttpRequestMessage httpRequestMessag
...
+                if (AuthenticationHelpers.IsATPoPSupported())
+                {
+                    GraphSession.Instance.GraphRequestProofofPossession.Request.Method = AuthenticationHelpers.ConvertToAzureRequestMethod(httpRequestMessage.Method);
+                    GraphSession.Instance.GraphRequestProofofPossession.Request.Uri.Reset(httpRequestMessage.RequestUri);
+                    foreach (var header in httpRequestMessage.Headers)
+                    {
+                        GraphSession.Instance.GraphRequestProofofPossession.Request.Headers.Add(header.Key, header.Value.First());
+                    }
+                    
+                    var accessToken = GraphSession.Instance.GraphRequestProofofPossession.BrowserCredential.GetTokenAsync(GraphSession.Instance.GraphRequestProofofPossession.PopTokenContext, cancellationToken).Result;
```
bgavrilMS, line 77: sync over async again.

src/Authentication/Authentication/Handlers/AuthenticationHandler.cs
```diff
@@ -87,6 +104,15 @@ private async Task<HttpResponseMessage> SendRetryAsync(HttpResponseMessage httpR
                 }
                 await DrainAsync(httpResponseMessage).ConfigureAwait(false);
 
+                if (AuthenticationHelpers.IsATPoPSupported())
+                {
+                    var popChallenge = httpResponseMessage.Headers.WwwAuthenticate.First(wa => wa.Scheme == "PoP");
```
bgavrilMS, line 109: You already parse this in the AuthenticationHelpers.cs class.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -86,6 +88,12 @@ private static bool IsWamSupported()
             return GraphSession.Instance.GraphOption.EnableWAMForMSGraph && SharedUtilities.IsWindowsPlatform();
         }
 
+        //Check to see if ATPoP is Supported
+        public static bool IsATPoPSupported()
```
bgavrilMS, line 92: I don't think this is correct. AT POP for public client (!) is supported when: 

- broker is available and enabled (currently this means Windows except Win Sever 2016 + WAM enabled)
- service supports AT POP

src/Authentication/Authentication.Core/Interfaces/IGraphRequestProofofPossession.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Uri Uri { get; set; }
+        HttpMethod HttpMethod { get; set; }
+        AccessToken AccessToken { get; set; }
+        string ProofofPossessionNonce { get; set; }
+        PopTokenRequestContext PopTokenContext { get; set; }
+        Request Request { get; set; }
+        InteractiveBrowserCredential BrowserCredential { get; set; }
```
bgavrilMS, line 20: Shouldn't this be `InteractiveBrokerCredential` ? Browser credential will not get you POP tokens.

src/Authentication/Authentication.Core/Interfaces/IGraphRequestProofofPossession.cs
bgavrilMS, line 12: This public interface is confusing me. There are no comments. It combines several things in it: 

- SHR details (uri, method) - part of the request
- access token - part of the response
- Azure Identity concept "credential" 
- Graph concepts.

I tend to think this is more of a https://en.wikipedia.org/wiki/God_object

docs/AT-Pop.md
```diff
@@ -0,0 +1,79 @@
...
+| Existing token, different URI | No new token, silent | New token, silent |
+| Existing expired token, below max token refreshes | New token, silent | New token, silent |
+| Existing expired token, exceeded max refreshes | New token, interactive | New token, interactive |
+
+## Installation
+
+To install the Microsoft Graph PowerShell SDK with the latest updates, use the following command:
+
+```powershell
+Install-Module -Name Microsoft.Graph -AllowClobber -Force
```
adamedx, line 36: We can leave out this section -- installation of this module is not changed by this feature.

docs/AT-Pop.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+Ensure you are using the latest version to access the AT PoP functionality.
+
+## Configuration
+
+### Enabling Access Token Proof of Possession
+
+To enable AT PoP, configure the Microsoft Graph SDK options as follows:
```
adamedx, line 45: If this option is specified, we should guarantee that we *only* use bound tokens, correct? We should fail if we can't get the bound token. Everyone agreed?

docs/AT-Pop.md
```diff
@@ -0,0 +1,79 @@
...
+
+```powershell
+Set-MgGraphOption -EnableATPoP $true
```
adamedx, line 48: nitpick: This should not be a boolean IMO.

adamedx, line 48: Should we release this specific feature as preview / experimental? Does MS Graph PS have this capability?

docs/authentication.md
```diff
@@ -112,6 +112,26 @@ When using `-AccessToken`, we won't have access to the refresh token and the cli
```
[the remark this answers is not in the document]
adamedx: > Having reviewed this PR, it has 2 very big problems:
> 
> * it is making un-authenticated request to graph every time a token is needed, to fetch a new nonce. **This will double the HTTP calls to Graph !** This is not needed - a new nonce should be retrieved from 200 OK response Authenticate-Info; Even if this is not available, it is better to reuse old nonce, it has a few min validity.
> * it is not taking into consideration that authenticated request with POP token can return 401 + invalid nonce response (e.g. nonce expired), in which case a new nonce must be retrieved.
> 
> For a prototype implementation of the protocol, please see https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PoPTokensProtocol/PoP_API_In_MSAL.md&_a=preview&version=GBdev&anchor=appendix---how-applications-can-use-this-lower-level-api

Need to check that the `Autheticate-Info` is supported by MS Graph.

docs/authentication.md
adamedx: I'm fine with this once we've proven the feature -- I'd like to avoid changing defaults until we've validated that we're not regressing or presenting user experience issues.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -113,23 +121,46 @@ private static async Task<InteractiveBrowserCredential> GetInteractiveBrowserCre
...
                     {
-                        // Run the thread in MTA.
-                        return interactiveBrowserCredential.Authenticate(new TokenRequestContext(authContext.Scopes), cancellationToken);
-                    });
+                        // Logic to implement ATPoP Authentication
+                        authRecord = await Task.Run(() =>
+                        {
+                            // Run the thread in MTA.
+                            //GraphSession.Instance.GraphRequestProofofPossession.AccessToken = interactiveBrowserCredential.GetTokenAsync(GraphSession.Instance.GraphRequestProofofPossession.PopTokenContext, cancellationToken).Result;
+                            return interactiveBrowserCredential.AuthenticateAsync(GraphSession.Instance.GraphRequestProofofPossession.PopTokenContext, cancellationToken);
```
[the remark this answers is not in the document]
FehintolaObafemi: No specific reason, would change this back for consistency

docs/AT-Pop.md
FehintolaObafemi: This is a good idea. @timayabi2020 can this be released as a preview version similar to the version 2.0 roll out?

docs/AT-Pop.md
FehintolaObafemi: Yes, that is the plan to only use bound tokens when this option is specified

src/Authentication/Authentication.Core/Interfaces/IGraphOptions.cs
FehintolaObafemi: Yes, that is the eventual plan. This option would be phased out eventually.

FehintolaObafemi: @peombwa Thoughts on this comment?

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
adamedx: The windows handle behavior isn't changing with this PR. That said, we are intentionally assuming that this *IS* a command-line app. The "MS Graph PowerShell SDK" is only meant to be used from within a PowerShell language session, and is not intended to interact with graphical applications. There is an MS Graph SDK .NET SDK that can be used for graphical applications. For a GUI app to use this "SDK," it would also need to load the PowerShell runtime as part of the process, which would not really make sense and is not done. I do think the name "SDK" is confusing here from a product standpoint, but that is more reflective of its auto-generated nature rather than it being used as an SDK for .net. It really is just limited to PowerShell's runtime / command-line environment.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
FehintolaObafemi: @bgavrilMS Is there a way to use the MSAL parser even when using Azure Identity?

bgavrilMS: Yes, it is a simple helper class. See https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.authenticationheaderparser?view=msal-dotnet-latest

See an example here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs#L444

docs/authentication.md
nkasco: Please consider #2790 before changing defaults.

src/Authentication/Authentication.Core/Interfaces/IGraphOptions.cs
darrelmiller: I don't think these settings should be part of an interface that has the very generic name IGraphOption.  All other authentication parameters are provided as parameters to the Connect-MgGraph. I don't understand why these properties should be handled differently.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +469,60 @@ public static Task DeleteAuthRecordAsync()
                 File.Delete(Constants.AuthRecordPath);
             return Task.CompletedTask;
         }
+
+        private static async Task<PopTokenRequestContext> CreatePopTokenRequestContext(IAuthContext authContext)
+        {
+            // Creating a httpclient that would handle all pop calls
+            Uri popResourceUri = GraphSession.Instance.GraphRequestPopContext.Uri ?? new Uri("https://graph.microsoft.com/beta/organization"); //PPE (https://graph.microsoft-ppe.com) or Canary (https://canary.graph.microsoft.com) or (https://20.190.132.47/beta/me)
```
darrelmiller, line 476: Is there no intention to support this for sovereign clouds?

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +469,60 @@ public static Task DeleteAuthRecordAsync()
...
+            GraphSession.Instance.GraphRequestPopContext.ProofofPossessionNonce = WwwAuthenticateParameters.CreateFromAuthenticationHeaders(popResponse.Headers, "Pop").Nonce;
+
+            // Refresh token logic --- start
+            var popPipelineOptions = new HttpPipelineOptions(new PopClientOptions()
+            {
+
+            });
+
+            var _popPipeline = HttpPipelineBuilder.Build(popPipelineOptions, new HttpPipelineTransportOptions());
+            GraphSession.Instance.GraphRequestPopContext.Request = _popPipeline.CreateRequest();
```
darrelmiller, line 491: How does this work in multi-threaded scenarios?

src/Authentication/Authentication.Core/Microsoft.Graph.Authentication.Core.csproj
```diff
@@ -1,19 +1,21 @@
...
   <PropertyGroup>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.11.4" />
-    <PackageReference Include="Azure.Identity.Broker" Version="1.1.0" />
-    <PackageReference Include="Microsoft.Graph.Core" Version="3.0.9" />
+    <PackageReference Include="Azure.Identity" Version="1.12.0-beta.3" />
+    <PackageReference Include="Azure.Identity.Broker" Version="1.2.0-beta.1" />
```
darrelmiller, line 15: In previous conversations, we had wondered if Azure.Identity is actually bringing any value here over using Microsoft.Identity.Client directly.  Azure.Identity is intended to limit the surface area exposed to 3P developers. This is not relevant for PowerShell scenarios.  I am especially concerned when I see the introduction of Azure.Core.Pipeline. We explicitly don't use the Azure Pipelining because it is only there to abstract away language differences, and again that isn't relevant because our PowerShell modules are only implemented in C#.

src/Authentication/Authentication/Handlers/AuthenticationHandler.cs
```diff
@@ -63,9 +74,24 @@ private async Task AuthenticateRequestAsync(HttpRequestMessage httpRequestMessag
...
             if (AuthenticationProvider != null)
             {
-                var accessToken = await AuthenticationProvider.GetAuthorizationTokenAsync(httpRequestMessage.RequestUri, additionalAuthenticationContext, cancellationToken: cancellationToken).ConfigureAwait(false);
-                if (!string.IsNullOrEmpty(accessToken))
-                    httpRequestMessage.Headers.Authorization = new AuthenticationHeaderValue(BearerAuthenticationScheme, accessToken);
+                if (GraphSession.Instance.GraphOption.EnableATPoPForMSGraph)
+                {
+                    GraphSession.Instance.GraphRequestPopContext.Request.Method = AuthenticationHelpers.ConvertToAzureRequestMethod(httpRequestMessage.Method);
+                    GraphSession.Instance.GraphRequestPopContext.Request.Uri.Reset(httpRequestMessage.RequestUri);
+                    foreach (var header in httpRequestMessage.Headers)
```
darrelmiller, line 81: How does this work when you do a parallel foreach?

docs/AT-Pop.md
timayabi2020: Yes, it can, however I think we should first publish the feature to an internal feed and get a few guys to test internally

src/Authentication/Authentication.Test/Microsoft.Graph.Authentication.Test.csproj
```diff
@@ -8,7 +8,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.2" />
     <!-- As described in this post https://devblogs.microsoft.com/powershell/depending-on-the-right-powershell-nuget-package-in-your-net-project, reference the SDK for dotnetcore-->
     <PackageReference Include="Microsoft.PowerShell.SDK" Version="7.2.2" PrivateAssets="all" Condition="'$(TargetFramework)' == 'net6.0'" />
-    <PackageReference Include="Moq" Version="4.20.69" />
+    <PackageReference Include="Moq" Version="4.20.1" />
```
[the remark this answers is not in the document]
FehintolaObafemi: Noted. Would be reverted

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -113,23 +116,43 @@ private static async Task<InteractiveBrowserCredential> GetInteractiveBrowserCre
...
+                            // Run the thread in MTA.
+                            return interactiveBrowserCredential.AuthenticateAsync(GraphSession.Instance.GraphRequestPopContext.PopTokenContext, cancellationToken);
+                        });
+                    }
+                    else
+                    {
+                        authRecord = await Task.Run(() =>
+                        {
+                            // Run the thread in MTA.
+                            return interactiveBrowserCredential.Authenticate(new TokenRequestContext(authContext.Scopes), cancellationToken);
```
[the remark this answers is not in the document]
FehintolaObafemi: The reason async overload is not being used here is to avoid MTA/STA issues. The initial implementation used async overload and we ran into issues with it. This goes back to the initial implementation of WAM for this sdk.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -446,5 +469,59 @@ public static Task DeleteAuthRecordAsync()
                 File.Delete(Constants.AuthRecordPath);
             return Task.CompletedTask;
         }
+
+        private static async Task<PopTokenRequestContext> CreatePopTokenRequestContext(IAuthContext authContext)
+        {
+            // Creating a httpclient that would handle all pop calls
+            Uri popResourceUri = GraphSession.Instance.GraphRequestPopContext.Uri ?? new Uri("https://graph.microsoft.com/beta/organization"); //PPE (https://graph.microsoft-ppe.com) or Canary (https://canary.graph.microsoft.com) or (https://20.190.132.47/beta/me)
+            HttpClient popHttpClient = new(new HttpClientHandler());
```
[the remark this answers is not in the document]
FehintolaObafemi: This function is called before GraphSession.Instance.GraphHttpClient has been called (assigned value for GraphSession.Instance.GraphHttpClient at this stage is null). A remainder that this is created for the sole purpose of getting the initial nonce value and is discarded afterwards to never be used again.

src/Authentication/Authentication.Core/Interfaces/IGraphOptions.cs
FehintolaObafemi: They are part of the interface which allows the user customize their msgraph ps experience. This is related to the Set-MgGraphOption and Get-MgGraphOption commands

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
FehintolaObafemi: The intention is to roll out this feature for all clouds

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
FehintolaObafemi: This has been resolved

src/Authentication/Authentication/Handlers/AuthenticationHandler.cs
FehintolaObafemi: This has been resolved

Issue #2476, opened 2023-12-05
Detect and generate correct content-type headers for input files


timayabi2020:
Currently for cmdlets that upload files, the media content-type headerhas been set to null for all file types. Therefore the API treats such requests as an application/octet-stream.
<img width="1267" alt="image" src="https://github.com/microsoftgraph/msgraph-sdk-powershell/assets/10947120/9543d59b-1d4c-4bec-9610-c8b3f9a65935">

There needs to be a feature that will automatically detect the file type and append the correct content-type header to the request.

Related issue is here on autorest  https://github.com/Azure/autorest.powershell/issues/1286

timayabi2020:
Closing this as it was fixed in https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/2570

PR #2542, opened 2024-01-24
Generate cmdlets from forked AutoREST.PowerShell repository.

### LINKED ISSUE #1570: Cmdlets should accept request headers
Some APIs such as https://docs.microsoft.com/en-us/graph/api/plannertaskdetails-update?view=graph-rest-1.0&tabs=http#request-headers require the caller to set a request header. Currently, there is no option for one to set request headers to the generated PS cmdlets when making a call.

ASK
Allow customers to pass a list of request headers as a parameter to the generated cmdlets.

timayabi2020:
Fixes #1570 and #2476

- Adds forked [Autorest ](https://github.com/microsoftgraph/autorest.powershell/tree/powershell-v2)as a submodule which will be used to generate cmdlets.
- Updates custom commands in the group module to include custom header parameter as expected by the auto generated Client API class.
- Updates pipelines to run generation using the forked AutoREST.PowerShell repo.
- Updates the global directive file (readme.graph.md) to point to the AutoREST.PowerShell submodule.

Generated cmdlets were published to an internal feed. Using cmdlets obtained from that feed manual tests involving PATCH, POST and DELETE operations were done while raptor handled GET operations. Use [this script](https://github.com/timayabi2020/AutorestFiles/blob/main/PowerShellAutorestFeed.ps1) to register the feed as a PS repository. 

**Sample cmdlet with optional custom header parameter**
image

**Sample cmdlet extracting the correct media type from file inputs and setting the correct request content type. (Intitally all input file requests defaulted to application/octet-stream
image

**Sample cmdlet with optional parameter for specifying the media/content type for file inputs**

image

**Sample raptor test run results for stable tests**
**NB** This was locally run because the feed was locally registered.
image

### DIFF
--- a/src/Groups/beta/custom/NewMgBetaGroupMember_Create.cs
+++ b/src/Groups/beta/custom/NewMgBetaGroupMember_Create.cs
@@ -43,7 +43,15 @@
 
         /// <summary>The reference to the client API class.</summary>
         public Groups Client => Module.Instance.ClientAPI;
+        /// <summary>Backing field for <see cref="CustomHeader" /> property.</summary>
+        private string _customHeader;
 
+        /// <summary>
+        /// CustomHeader Parameter. This should have a key:value and comma separated for multiple key:value pairs
+        /// </summary>
+        [System.Management.Automation.Parameter(Mandatory = false, HelpMessage = "CustomHeader Parameter. This should have a key:value and comma separated for multiple key:value pairs", ValueFromPipeline = true)]
+        [Category(ParameterCategory.Runtime)]
+        public string CustomHeader { get => this._customHeader; set => this._customHeader = value; }
         /// <summary>Backing field for <see cref="GroupId" /> property.</summary>
         private string _groupId;
 
@@ -262,12 +270,12 @@
                 try
                 {
                     await ((Runtime.IEventListener)this).Signal(Runtime.Events.CmdletBeforeAPICall); if (((Runtime.IEventListener)this).Token.IsCancellationRequested) { return; }
-                    await this.Client.GroupCreateMemberGraphBPreRef(GroupId, BodyParameter, onNoContent, onDefault, this, Pipeline);
+                    await this.Client.GroupCreateMemberGraphBPreRef(GroupId, CustomHeader, BodyParameter, onNoContent, onDefault, this, Pipeline);
                     await ((Runtime.IEventListener)this).Signal(Runtime.Events.CmdletAfterAPICall); if (((Runtime.IEventListener)this).Token.IsCancellationRequested) { return; }
                 }
                 catch (Microsoft.Graph.Beta.PowerShell.Runtime.UndeclaredResponseException urexception)
                 {
-                    WriteError(new global::System.Management.Automation.ErrorRecord(urexception, urexception.StatusCode.ToString(), global::System.Management.Automation.ErrorCategory.InvalidOperation, new { GroupId = GroupId, body = BodyParameter })
+                    WriteError(new global::System.Management.Automation.ErrorRecord(urexception, urexception.StatusCode.ToString(), global::System.Management.Automation.ErrorCategory.InvalidOperation, new { GroupId = GroupId, CustomHeader = CustomHeader, body = BodyParameter })
                     {
                         ErrorDetails = new global::System.Management.Automation.ErrorDetails(urexception.Message) { RecommendedAction = urexception.Action }
                     });
--- a/.azure-pipelines/common-templates/checkout.yml
+++ b/.azure-pipelines/common-templates/checkout.yml
@@ -10,7 +10,7 @@ steps:
   - checkout: self
     clean: true
     fetchDepth: 1
-    submodules: true
+    submodules: recursive
     persistCredentials: true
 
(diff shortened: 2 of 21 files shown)


calebkiage:
Is it possible to use the same syntax as CLI for headers?

i.e. `--headers AcceptLanguage=en`

timayabi2020:
@calebkiage @sebastienlevert yes it can change, however I preferred the approach of setting http headers the way it is normally done in a curl request. e.g ``curl -H "Accept-Charset: utf-8" -H "Content-Type: application/json" https://graph.com.........``

sebastienlevert:
I think we should be using the same approach that Invoke-WebRequest (and other similar Cmdlets) are using. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.4#-headers

src/Groups/beta/custom/NewMgBetaGroupMember_Create.cs
```diff
@@ -43,7 +43,15 @@ public partial class NewMgBetaGroupMember_Create : System.Management.Automation.
...
         public Groups Client => Module.Instance.ClientAPI;
+        /// <summary>Backing field for <see cref="CustomHeader" /> property.</summary>
+        private string _customHeader;
 
+        /// <summary>
+        /// CustomHeader Parameter. This should have a key:value and comma separated for multiple key:value pairs
+        /// </summary>
+        [System.Management.Automation.Parameter(Mandatory = false, HelpMessage = "CustomHeader Parameter. This should have a key:value and comma separated for multiple key:value pairs", ValueFromPipeline = true)]
+        [Category(ParameterCategory.Runtime)]
+        public string CustomHeader { get => this._customHeader; set => this._customHeader = value; }
```
peombwa, line 54: Can we:
- [ ] Change the `CustomHeader` type to a `IDictionary`. This will allow customers to use familiar PowerShell syntax when setting headers. See [Invoke-MgGraphRequest](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/a00ce42fc41d6c196e18e14d2649857f3feb41c4/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs#L72) implementation and [invoke-restmethod](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7.4#example-5-pass-multiple-headers) example.
- [ ] Rename `CustomHeader` parameters to `CustomHeaders` since customers can set more than one header value. This will align with the existing naming convention in `Invoke-MgGraphRequest` and other HTTP based commands.

The suggestions above will need to be applied to the AutoREST fork as well.

peombwa (changes requested):
Thanks for addressing on the most asked for feature! Please see my comment below on improvements to the UX of the parameter.

src/Groups/beta/custom/NewMgBetaGroupMember_Create.cs
peombwa: The final experience should something like:
``` Powershell
# Setting a single header.
Get-MgUser -UserId $userId -CustomHeaders @{ 'Accept-Language' = 'en' }

# Setting multiple headers (option 1).
Get-MgUser -UserId $userId -CustomHeaders @{ 'Accept-Language' = 'en'; 'MySpecialHeader' = 'val1' }

# Setting multiple headers (option 2)
$MyCustomHeaders = @{
   'Accept-Language' = 'en'
   'MySpecialHeader' = 'val1'
}
Get-MgUser -UserId $userId -CustomHeaders $MyCustomHeaders 
```

We should also handle cases where duplicate headers are set.

timayabi2020: @calebkiage @sebastienlevert @peombwa thanks for the suggestion I will go ahead and implement the changes suggested by Peter

calebkiage: @timayabi2020, @peombwa I still think the name should be `Headers` and not `CustomHeaders`, the `Invoke-WebRequest` cmdlet also uses that naming and it's shorter to type.

calebkiage: @sebastienlevert, what do you think?

timayabi2020: @calebkiage we are using ``CustomHeaders`` which communicates that there is an option for customers to provide additional information that may be pertinent to them.  It will just append the information to the already existing default headers sent to graph APIs.

sebastienlevert: In all other SDKs we use Headers in our request configurations to append to existing ones set by the underlying platform. +1 to naming it Headers.

timayabi2020: image

timayabi2020: Ok ``Headers`` it is.
image

peombwa (approved):
:shipit:

PR #2574, opened 2024-02-09
Patch Release. Ensures that ``-Header`` parameter is not mandatory

timayabi2020:
Currently for all cmdlets one must specify the ``-Header`` parameter.
Updated AutoREST code generator to prevent this scenario.


calebkiage:
I think the submodule needs to be updated.

calebkiage (approved):

PR #2615, opened 2024-02-27
Add new app id

ElinorW:
﻿

- Adding the new application Id for the Torus migration
-
-

### Other links

-
-
-


ElinorW:
no longer needed

PR #2680, opened 2024-04-15
Orders permissions based on least privilege property value (``True`` or ``false``)

### LINKED ISSUE #2618: Find-MgGraphCommand sorts required permissions alphabetically, not least privilege
**Describe the bug**
For a solution I am working on, I am using the Graph via the PowerShell Graph modules. In order to determine the required permissions, I am trying the Find-MgGraphCommand cmdlet to find the most restrictive permissions. Unfortunately, that cmdlet has the issue that permissions are not ordered based on permission level, but it looks like they are sorted alphabetically (or at least partially). That makes the cmdlet much less useful.

**To Reproduce**
1. Run `(Find-MgGraphCommand -Command Get-MgOrganization)[0].Permissions`. This command returns a list that is sorted alphabetically: 
image

In the documentation, the list is ordered based on least privilege: [Get organization](https://learn.microsoft.com/en-us/graph/api/organization-get?view=graph-rest-1.0&tabs=http)
image

**Expected behavior**
The expected behavior is that the sorting is based on least to most privileged, which makes much more sense.

**Debug Output**
> Run the problematic command with `-Debug` and paste the resulting debug stream below.
> ⚠ ATTENTION: Be sure to remove any sensitive information that may be in the logs.

**Module Version**
2.14.1

**Environment Data**
PSVersion                      5.1.22621.2506
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.22621.2506
CLRVersion                     4.0.30319.42000

**Screenshots**
> If applicable, add screenshots to help explain your problem.

**Additional context**
> Add any other context about the problem here.

timayabi2020:
﻿

Fixes #2618 

- Updated NewCommandMetadata.ps1 to order permissions from graphexplorer api based on the least privilege property. (See image below)
- Updated GraphCommand.cs by adding ``IsLeastPrivilege`` property

image

### DIFF
--- a/src/Authentication/Authentication/Models/GraphCommand.cs
+++ b/src/Authentication/Authentication/Models/GraphCommand.cs
@@ -103,5 +103,9 @@ namespace Microsoft.Graph.PowerShell.Authentication.Models
         /// Full deescription of the permission.
         /// </summary>
         public string FullDescription { get; set; }
+        /// <summary>
+        /// Least privilege flag shows whether the permission is the least privilege.
+        /// </summary>
+        public bool IsLeastPrivilege { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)


andrueastman (approved):

ykuijs:
The sorting looks much better, but there is still one thing that is a bit odd:
The order now lists Organization.ReadWrite.All before Organization.Read.All. I think that should be the other way around.

PR #2690, opened 2024-04-18
Restricts escaping of special characters to path segments only when using Invoke-MgGraphRequest and bumps ``Azure.Identity.Broker`` Library

### LINKED ISSUE #1947: Invoke-MgGraphRequest can't handle URI with #

timayabi2020:
Fixes #1947 and ``Azure.Identity.Broker`` package conflict due to bumping of Azure Identity [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/2675)

- Restricted escaping/URL encoding of special characters in path segments only. Uri with system query options should not URL encoded values. 

**Uri with OData system query options**
image

**Uri with special characters in path segments**
image

- Updated ``Azure.Identity.Broker`` due to interactive browser errors  (``Method 'get_UseDefaultBrokerAccount' in type 'Azure.Identity.Broker.InteractiveBrowserCredentialBrokerOptions' from assembly 'Azure.Identity.Broker, Version=1.0.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8' does not have an implementation``) as a result of bumping the ``Azure.Identity``. The version bump is based on the library documentation [here](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/identity.broker-readme?view=azure-dotnet)

### DIFF
--- a/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
+++ b/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
@@ -15,17 +15,24 @@ namespace Microsoft.Graph.PowerShell.Authentication.Helpers
     {
         public static Uri EscapeDataStrings(this Uri uri)
         {
-            if(uri.Query.Length > 0)
-                return uri;
             int counter = 0;
             var pathSegments = uri.OriginalString.Split('/');
             StringBuilder sb = new StringBuilder();
-            foreach (var segment in pathSegments)
+            foreach (var s in pathSegments)
             {
+                var segment = s;
                 //Skips the left part of the uri i.e https://graph.microsoft.com
                 if (counter > 2)
                 {
                     sb.Append('/');
+                    if(s.Contains("?"))
+                    {
+                        var queryStringIndex = segment.IndexOf("?");
+                        string queryString = segment.Substring(queryStringIndex);
+                        segment = s.Substring(0, queryStringIndex);
+                        sb.Append(Uri.EscapeDataString(segment));
+                        sb.Append(queryString);
+                    }
                     sb.Append(Uri.EscapeDataString(segment));
                 }
                 counter++;
--- a/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
+++ b/src/Authentication/Authentication/Helpers/EncodeUriPath.cs
@@ -33,7 +33,10 @@ namespace Microsoft.Graph.PowerShell.Authentication.Helpers
                         sb.Append(Uri.EscapeDataString(segment));
                         sb.Append(queryString);
                     }
-                    sb.Append(Uri.EscapeDataString(segment));
+                    else
+                    {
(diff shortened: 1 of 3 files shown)


src/Authentication/Authentication/Helpers/EncodeUriPath.cs
```diff
@@ -15,6 +15,8 @@ public static class EncodeUriPaths
     {
         public static Uri EscapeDataStrings(this Uri uri)
         {
+            if(uri.Query.Length > 0)
```
andrueastman, line 18: The only question I have is what happens when the Uri has both path segments and query parameters?
e.g. `https://graph.microsoft.com/v1.0/users/userswith#character?$select=id,displayName`

timayabi2020: Nice catch!. I've made corrections to address this.

src/Authentication/Authentication/Helpers/EncodeUriPath.cs
```diff
@@ -18,12 +18,21 @@ public static Uri EscapeDataStrings(this Uri uri)
...
                 {
                     sb.Append('/');
+                    if(s.Contains("?"))
+                    {
+                        var queryStringIndex = segment.IndexOf("?");
+                        string queryString = segment.Substring(queryStringIndex);
+                        segment = s.Substring(0, queryStringIndex);
+                        sb.Append(Uri.EscapeDataString(segment));
+                        sb.Append(queryString);
+                    }
```
andrueastman, line 35: Do you need an `else` block here? Looks  the appending will happen twice?

timayabi2020: Yes added

andrueastman (approved):
👍🏼

PR #2699, opened 2024-04-24
Add flag to send certificate chain for subject name / issuer based auth in Connect-MgGraph

### LINKED ISSUE #2698: Support for Subject Name / Issuer based certificate authentication
### Is your feature request related to a problem? Please describe the problem.

I am trying to use a certificate with SN+I (subject name / issuer) based authentication which requires sending entire certificate chain when requesting an access token. But `Connect-MgGraph` doesn't have an option to enable that behavior. 

### Describe the solution you'd like.

Add a flag to `Connect-MgGraph` to allow sending entire certificate chain while requesting an access token. This option is already available in Azure Identity SDK so just needs to be plumbed through the cmd-lets. 

### Additional context?

_No response_

nayanshah:
﻿

Fixes #2698 

Adds a flag `SendCertificateChain` to `Connect-MgGraph` to send entire certificate chain while requesting access token. This allows subject name / issuer based auth.

### Other links

`ClientCertificateCredentialOptions.SendCertificateChain` - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/ClientCertificateCredentialOptions.cs

### DIFF
--- a/src/Authentication/Authentication.Core/Interfaces/IAuthContext.cs
+++ b/src/Authentication/Authentication.Core/Interfaces/IAuthContext.cs
@@ -47,6 +47,7 @@ namespace Microsoft.Graph.PowerShell.Authentication
         string Account { get; set; }
         string CertificateThumbprint { get; set; }
         string CertificateSubjectName { get; set; }
+        bool SendCertificateChain {  get; set; }
         X509Certificate2 Certificate { get; set; }
         ContextScope ContextScope { get; set; }
         Version PSHostVersion { get; set; }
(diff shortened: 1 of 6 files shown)


timayabi2020 (changes requested):
I recommend that you add a pester test for the new parameter.

nayanshah:
I don't see any test passing a real certificate or a way to mock the authentication provider, so would be difficult to write a test for the new parameter. The parameter is forwarded to `ClientCertificateCredential` class which doesn't expose the value as a public member, so not straightforward to add a C# unit test either. 

I have updated existing pester test to include the new parameter in the appropriate parameter set. Would that be sufficient given that not much new logic is being added?

nayanshah:
Thank you @timayabi2020 . Do you know when is the next release scheduled for?

timayabi2020:
@nayanshah it will be out by end of this week.

PR #2773, opened 2024-06-03
Fix Metadata breaking change workflow

Ndiritu:
This PR:
- reduces unnecessary runs to those that affect generation config files
- gives job write permissions to fix current failures e.g. https://github.com/microsoftgraph/msgraph-sdk-powershell/actions/runs/9353079967
- updates workflow to fail if OpenAPI errors exist & creates a PR review to prevent merging with breaking changes.


andrueastman (approved):
👍🏼

PR #2833, opened 2024-07-08
Upgrades moq version to 4.20.70

timayabi2020:
This PR is part of a security compliance approach in response to data exfiltration issues during Moq installation as reported [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2090).


andrueastman:
Should we also update the config here for dependabot so that this captured in fiture updates?
https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/.github/dependabot.yml

PR #2865, opened 2024-07-26
[v2] Examples Update

timayabi2020:
This pull request was automatically created by Azure Pipelines. **Important** Check for unexpected deletions or changes in this PR.


andrueastman (approved):

timayabi2020:
Not merging this because the deleted example is correct as per the API reference here https://learn.microsoft.com/en-us/graph/api/synchronization-synchronization-list-templates?view=graph-rest-beta&tabs=powershell. 
The example import script needs to be reviewed

PR #2971, opened 2024-09-26
Fixes subsequent failures after first successful ``Invoke-MgGraphRequest`` run when using national cloud environments

### LINKED ISSUE #2950: Invoke-MGGraphRequest fails after first call as of v2.21.0 with USGov Environment
### Describe the bug

When attempting to use Invoke-MGGraphRequest against the USGov environment, the first call succeeds however subsequent calls fail. The first attempt correctly goes to graph.microsoft.us however subsequent calls are sent to graph.microsoft.com which results in an authentication failure. This appears to be caused by https://github.com/microsoftgraph/msgraph-sdk-powershell/commit/815b63f7d6ff529880bf12441272850a35d2f293 

### Expected behavior

Subsequent Invoke-MGGraphRequest calls continue functioning after the initial call.

### How to reproduce

```Import-Module Microsoft.Graph.Authentication -MinimumVersion 2.21.0```
```Connect-MgGraph -Environment "USGov" -Scopes "Directory.AccessAsUser.All"```
Works: ```Invoke-MgGraphRequest -Uri "v1.0/organization"```
Fails: ```Invoke-MgGraphRequest -Uri "v1.0/me"```
Fails: ```Invoke-MgGraphRequest -Uri "v1.0/organization"```

### SDK Version

2.21.0 and later

2.21

### Known Workarounds

Use version 2.20.0 or earlier of module Microsoft.Graph.Authentication

### Debug output

PS C:\> Invoke-MgGraphRequest -Uri "v1.0/organization" -Debug
VERBOSE: GET https://graph.microsoft.com/v1.0/organization with 0-byte payload

Confirm
Continue with this operation?
[Y] Yes  [A] Yes to All  [H] Halt Command  [S] Suspend  [?] Help (default is "Y"): a
DEBUG: GET /v1.0/organization HTTP/1.1
HTTP: graph.microsoft.com
User-Agent: Mozilla/5.0 (Windows NT 10.0; Microsoft Windows 10.0.22631; en-US) PowerShell/7.4.5 Invoke-MgGraphRequest

VERBOSE: received 17413-byte response of content type application/json
DEBUG: GET https://graph.microsoft.us/v1.0/organization
HTTP/2.0 200 OK
Cache-Control: no-cache
Vary: Accept-Encoding
Strict-Transport-Security: max-age=31536000
request-id: fbfc3588-61a7-48d6-bf58-352c72839120
client-request-id: 42c76821-de22-44f3-8496-0b5bdc35ea53
x-ms-ags-diagnostic: {"ServerInfo":{"DataCenter":"USGov Texas","Slice":"E","Ring":"4","ScaleUnit":"001","RoleInstance":"SN5NEPF00001F93"}}
odata-version: 4.0
Date: Mon, 09 Sep 2024 19:14:04 GMT
Content-Type: application/json; odata.metadata=minimal; odata.streaming=true; IEEE754Compatible=false; charset=utf-8
Content-Encoding: gzip

{"@odata.context":"https://graph.microsoft.us/v1.0/$metadata#organization","value":"redacted"}

@odata.context                 https://graph.microsoft.us/v1.0/$metadata#organization
value                          {redacted}

PS C:\> Invoke-MgGraphRequest -Uri "v1.0/me" -Debug
VERBOSE: GET https://graph.microsoft.com/v1.0/me with 0-byte payload

Confirm
Continue with this operation?
[Y] Yes  [A] Yes to All  [H] Halt Command  [S] Suspend  [?] Help (default is "Y"): a
DEBUG: GET /v1.0/me HTTP/1.1
HTTP: graph.microsoft.com
User-Agent: Mozilla/5.0 (Windows NT 10.0; Microsoft Windows 10.0.22631; en-US) PowerShell/7.4.5 Invoke-MgGraphRequest

Invoke-MgGraphRequest: Unable to perform redirect as Location Header is not set in response
PS C:\> Invoke-MgGraphRequest -Uri "v1.0/organization" -Debug
VERBOSE: GET https://graph.microsoft.com/v1.0/organization with 0-byte payload

Confirm
Continue with this operation?
[Y] Yes  [A] Yes to All  [H] Halt Command  [S] Suspend  [?] Help (default is "Y"): a
DEBUG: GET /v1.0/organization HTTP/1.1
HTTP: graph.microsoft.com
User-Agent: Mozilla/5.0 (Windows NT 10.0; Microsoft Windows 10.0.22631; en-US) PowerShell/7.4.5 Invoke-MgGraphRequest

Invoke-MgGraphRequest: Unable to perform redirect as Location Header is not set in response

### Configuration

PSVersion                      7.4.5
PSEdition                      Core
GitCommitId                    7.4.5
OS                             Microsoft Windows 10.0.22631
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}

### Other information

_No response_

### PR DESCRIPTION
**Pull request by timayabi2020**
Fixes #2950 
The fix provided [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/commit/815b63f7d6ff529880bf12441272850a35d2f293) only works once when an absolute uri is passed to ``Invoke-MgGraphRequest`` cmdlet after connecting to a national cloud environment.  Subsequent requests after the first successful one fails.

-  Added Microsoft graph context check in ``Find-MgGraphPermission`` in order to get the correct environment and extract the correct graph endpoint instead of using the default graph one (https://graph.microsoft.com) assigned to global environment. 
- Reset GraphSession environment back to its original state defined by the user.

### Test scenarios.
- Check that ``Find-MgGraphPermission`` does not default the session to the global graph endpoint after connecting to a national cloud environment.
- Subsequent calls to graph endpoints do not fail after connecting to a national cloud environment. 

NB: See attached image below for the above tests
image

### DIFF
--- a/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
+++ b/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
@@ -1004,7 +1004,6 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
         private void ResetGraphSessionEnvironment()
         {
             _originalEnvironment = GraphSession.Instance.Environment;
-            GraphSession.Instance.Environment = _originalEnvironment;
         }
 
         #region CmdLet LifeCycle
--- a/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
+++ b/src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
@@ -1003,11 +1003,8 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
         /// </summary>
         private void ResetGraphSessionEnvironment()
         {
-            var currentEnvironment = GraphSession.Instance.Environment;
-            if(currentEnvironment != null && !currentEnvironment.Equals(_originalEnvironment))
-            {
-                GraphSession.Instance.Environment = _originalEnvironment;
-            }
+            _originalEnvironment = GraphSession.Instance.Environment;
+            GraphSession.Instance.Environment = _originalEnvironment;
         }
 
         #region CmdLet LifeCycle
@@ -1046,8 +1043,6 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
                             if (ShouldCheckHttpStatus && !isSuccess)
                             {
                                 var httpErrorRecord = await GenerateHttpErrorRecordAsync(httpResponseMessageFormatter, httpRequestMessage);
-                                // A reset of the GraphSession Environment is required to avoid side effects
-                                ResetGraphSessionEnvironment();
                                 ThrowTerminatingError(httpErrorRecord);
                             }
                             await ProcessResponseAsync(httpResponseMessage);
--- a/src/Authentication/Authentication/custom/common/Permissions.ps1
+++ b/src/Authentication/Authentication/custom/common/Permissions.ps1
@@ -34,7 +34,20 @@ function Permissions_GetPermissionsData([bool] $online) {
     # Make a REST request to MS Graph to get the permissions data from the Microsoft Graph service principal
     if ( $online -or ! $_permissions.msGraphServicePrincipal -or ! $_permissions.isFromInvokeMgGraphRequest ) {
         try {
-            $restResult = Invoke-MgGraphRequest -method GET -OutputType PSObject $_permissions.msGraphPermissionsRequestUri
+            
+            # Get-MgContext is used to get the current context for the request to MS Graph
+            # From the context, we can get the current environment and use it to get the permissions request URI
+            # If the context is not available, then we will use the default permissions request URI
+
+            $context = Get-MgContext
+            $uri = $_permissions.msGraphPermissionsRequestUri
+            if($context){
+                $currentEnv = $context.Environment
+                $allEnv = Get-MgEnvironment
+                $env = $allEnv | Where-Object { $_.Name -eq $currentEnv }
+                $uri = $env.GraphEndpoint + "/v1.0/servicePrincipals?`$filter=appId eq '$Permissions_msGraphApplicationId'"
+            }
+            $restResult = Invoke-MgGraphRequest -method GET -OutputType PSObject $uri
 
             if ( $restResult ) {
                 $_permissions.msGraphServicePrincipal = $restResult | Select-Object -ExpandProperty value

### CODE REVIEW

File: src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs [hunk 1]
```diff
@@ -1003,11 +1003,8 @@ private string QualifyFilePath(string path)
         /// </summary>
         private void ResetGraphSessionEnvironment()
         {
-            var currentEnvironment = GraphSession.Instance.Environment;
-            if(currentEnvironment != null && !currentEnvironment.Equals(_originalEnvironment))
-            {
-                GraphSession.Instance.Environment = _originalEnvironment;
-            }
+            _originalEnvironment = GraphSession.Instance.Environment;
+            GraphSession.Instance.Environment = _originalEnvironment;
```
Ndiritu, line 1007: The values will already be the same?

src/Authentication/Authentication/custom/common/Permissions.ps1
```diff
@@ -34,7 +34,20 @@ function Permissions_GetPermissionsData([bool] $online) {
...
+            # If the context is not available, then we will use the default permissions request URI
+
+            $context = Get-MgContext
+            $uri = $_permissions.msGraphPermissionsRequestUri
+            if($context){
+                $currentEnv = $context.Environment
+                $allEnv = Get-MgEnvironment
+                $env = $allEnv | Where-Object { $_.Name -eq $currentEnv }
+                $uri = $env.GraphEndpoint + "/v1.0/servicePrincipals?`$filter=appId eq '$Permissions_msGraphApplicationId'"
+            }
```
Ndiritu, line 49: Not sure but would it be better if this happens under `_Permissions_Initialize`? Esp in case the other methods also need to call Graph

timayabi2020: When you import authentication module into the session i.e. ``Import-Module Microsoft.Graph.Authentication``, you are loading ``Find-MgGraphPermissions`` cmdlet which [loads](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/fddf88ca79b9dcefb51cee1f9a4d22e0df5e94d6/src/Authentication/Authentication/custom/Find-MgGraphPermission.ps1#L7) the ``Permissions.ps1`` script. Therefore, having that block in that function will result to null stack traces thrown in a PowerShell session because at that point a connection to graph has not yet been done. i.e. Null graph context in the session. Keep in mind there are some users who will have the authentication import step and may have enforced strict error handling in their scripts. In short that code block only comes into play when a connection to graph has been established and the context set in the session.

src/Authentication/Authentication/Cmdlets/InvokeMgGraphRequest.cs
timayabi2020: ``_originalEnvironment`` holds the previous state defined by the user. So to ensure subsequent requests by other cmdlets don't end up calling the default Global graph endpoint especially when connected to a national cloud, we assign the GraphSession.Instance.Environment to the value the `_originalEnvironment`` holds even after we are done using ``Invoke-MgGraphCommand``

Ndiritu: Thanks for sharing context. Still looks like we may not need one of the two lines since both variables will have the same value after line `1006`.

Ndiritu (approved):

PR #2979, opened 2024-10-01
Update MgCommandMetadata.json

lukasgd1913:
Added permissionreference for "/identity/conditionalAccess" entries.

﻿

Fixes #

-
-
-

### Other links

-
-
-


timayabi2020:
Hi @lukasgd1913 thank you for making this contribution. Kindly note that ``MgCommandMetadata.json`` is autogenerated during cmdlet build. You can open the issue here https://github.com/microsoftgraph/msgraph-sdk-powershell/issues so that the team responsible can be tagged and respond to it.

lukasgd1913:
thanks for the message @timayabi2020 
I am aware that this is autogenerated. This was already in the description of the pull request. But it also said that you look at the pull request and adjust it in the automation.

timayabi2020:
@lukasgd1913 for every WeeklyDocsRefresh PR which is responsible for updating the Open Api files used for generating/updating cmdlets, the ``MgCommandMetadata.json`` file is also updated with metadata for generated cmdlets.  This means that even if we check in your changes they will still be overwritten in the next automated run.  My suggestion is that we should try and solve the issue you are addressing in a different way, other than directly modifying the metadata file. i.e.  Find out why we have missing permission references for some paths.

Also, if you look at the PRs responsible for updating metadata file, the message is that we should check for unnecessary deletions and additions and not go ahead to directly modify anything on the generated files.  This is supposed to prompt the reviewer to check on the process responsible for making the changes.

PR #2995, opened 2024-10-15
2989 migrate ado release pipeline from classic UI to 1es

timayabi2020:
﻿

Fixes #2989 

### Changes proposed in this pull request

- Added stages for publishing modules to an internal feed and PowerShell gallery.
- Configured approval step for publishing to PowerShell gallery (See image below)
image

PR #2998, opened 2024-10-17
Adds condition to deploy to ps gallery on main branch triggers only.

timayabi2020:
This is a follow up to the merged [PR](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/2995) to only do PS Gallery deployments via main branch. The stage should be skipped for other branches.

image


andrueastman (approved):
👍🏼

PR #2999, opened 2024-10-17
Handles ``+`` character in supported equality and function operators in filter expressions

### LINKED ISSUE #2973: Filter is incorrectly handling special characters
### Describe the bug

When attempting to use a + symbol in a filter string the resulting query is not correctly encoded

### Expected behavior

+ symbol is encoded as %2B in the request made against graph 

### How to reproduce

1. run `Get-MgUser -Filter "onPremisesImmutableId in ('AegKW3DO+Uin3xHz7t0pIw==', 'YQuy6DbntUCGREZeTT5YwA==')" -Debug`
2. See that the request made contains a space in the value supplied as the $filter query parameter. 

```log
HTTP Method:
GET

Absolute Uri:
https://graph.microsoft.com/v1.0/users?$filter=onPremisesImmutableId in %28%27AegKW3DO Uin3xHz7t0pIw%3D%3D%27%2C %27YQuy6DbntUCGREZeTT5YwA%3D%3D%27%29
```

### SDK Version

2.23.0

_No response_

### Known Workarounds

```PowerShell
Invoke-MgRestMethod -Uri 'https://graph.microsoft.com/v1.0/users?$filter=onPremisesImmutableId%20in%20%28%27AegKW3DO%2BUin3xHz7t0pIw%3D%3D%27%2C%20%27YQuy6DbntUCGREZeTT5YwA%3D%3D%27%29' -Debug
```

results in the correct encoding of the URL parameters:
```log
DEBUG: GET /v1.0/users?$filter=onPremisesImmutableId%20in%20%28%27AegKW3DO%2BUin3xHz7t0pIw%3D%3D%27%2C%20%27YQuy6DbntUCGREZeTT5YwA%3D%3D%27%29 HTTP/1.1
HTTP: graph.microsoft.com
```

### Debug output

<details><summary>Click to expand log</summary>

```
DEBUG: [CmdletBeginProcessing]: - Get-MgUser begin processing with parameterSet 'List'.
DEBUG: [Authentication]: - AuthType: 'Delegated', TokenCredentialType: 'DeviceCode', ContextScope: 'CurrentUser', AppName: 'Microsoft Graph Command Line Tools'.
DEBUG: [Authentication]: - Scopes: [Application.ReadWrite.All, AppRoleAssignment.ReadWrite.All, AuditLog.Read.All, Calendars.ReadWrite, openid, profile, User.Read, User.Read.All, User.ReadWrite, email].
DEBUG: ============================ HTTP REQUEST ============================

HTTP Method:
GET

Absolute Uri:
https://graph.microsoft.com/v1.0/users?$filter=onPremisesImmutableId in %28%27AegKW3DO Uin3xHz7t0pIw%3D%3D%27%2C %27YQuy6DbntUCGREZeTT5YwA%3D%3D%27%29

Headers:
FeatureFlag                   : 00000043
Cache-Control                 : no-store, no-cache
User-Agent                    : Mozilla/5.0,(Windows NT 10.0; Microsoft Windows 10.0.26100; en-US),PowerShell/7.4.5
Accept-Encoding               : gzip
SdkVersion                    : graph-powershell/2.23.0
client-request-id             : cab71d36-315b-4ddc-ac84-e62dd819d10b

Body:



DEBUG: ============================ HTTP RESPONSE ============================

Status Code:
OK

Headers:
Cache-Control                 : no-cache
Transfer-Encoding             : chunked
Vary                          : Accept-Encoding
Strict-Transport-Security     : max-age=31536000
request-id                    : ff1f401e-48c6-4226-bdf3-06508749b56e
client-request-id             : cab71d36-315b-4ddc-ac84-e62dd819d10b
x-ms-ags-diagnostic           : {"ServerInfo":{"DataCenter":"West US 2","Slice":"E","Ring":"4","ScaleUnit":"003","RoleInstance":"CO1PEPF00004BE8"}}
x-ms-resource-unit            : 2
OData-Version                 : 4.0
Date                          : Fri, 27 Sep 2024 20:48:23 GMT

Body:
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users",
  "value": []
}


DEBUG: [CmdletEndProcessing]: - Get-MgUser end processing.
```
</details>

### Configuration

PS C:\Users\gavinbarron> $PSVersionTable

PSVersion                      7.4.5
PSEdition                      Core
GitCommitId                    7.4.5
OS                             Microsoft Windows 10.0.26100
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}

### Other information

This issue is being tracked as an IcM  id=544584306

timayabi2020:
﻿

Fixes #2973

- This change handles scenarios where equality operator types like ``in`` and ``has`` contain ``+`` character enclosed within brackets in filter expressions.
- It also handles scenarios where function operator types like ``startsWith`` contain ``+`` characters). 

image


tools/Custom/PSCmdletExtensions.cs
```diff
@@ -29,7 +29,7 @@ internal static string UnescapeString(this PSCmdlet cmdlet, string value)
             try
             {
                 var unescapedString = Uri.UnescapeDataString(value);
-                return value.EndsWith("'") ? unescapedString: unescapedString.Replace('+', ' ');
+                return (value.EndsWith("'") || value.EndsWith("')")) ? unescapedString: unescapedString.Replace('+', ' ');
```
andrueastman, line 32: Is it possible to capture these scenarios in a test to prevent a regression in the future?

timayabi2020: This will be tricky because the extension method a cmdlet derived from ``PSCmdlet`` class and cmdlets derived from the PSCmdlet class in PowerShell cannot be invoked directly because they are designed to be used within the context of a PowerShell runtime environment.

For example in a unit test we will need to have something like the code below calling a cmdlet implementation.
```
        var cmdlet = new CommandA();
        cmdlet.Value = "('Khalayi+mzuri')";
        cmdlet.Invoke();
```
The above won't work because we will not have a PowerShell run time which ``PSmdlet`` class depends on.

Below is the cmdlet implementation which we hope it can use the extension method.
```
using System.Management.Automation;

public class CommandA : PSCmdlet
{
    [Parameter(Position = 0, Mandatory = true)]
    public string Value { get; set; }

    protected override void ProcessRecord()
    {
       WriteObject(this.UnescapeString(Value));
    }
}
```
This is one of the limitations we currently having when it comes to writing unit testing some an extension method expecting a Cmdlet class. 

Maybe you can give more suggestion on how to go about this.

andrueastman: Thanks for extra context here @timayabi2020 

I believe it should still be possible to add a unit test library to the `tools` projects and have the unit tests target the specific method. Another alternative would be to refactor out the extension methods into separately testable components so that these methods simply call the testable methods. (All these are ideas thoughts outside the scope of this PR though, would you be willing to create a separate issue to investigate?)

Otherwise, we can resolve this on for now.

andrueastman (approved):

PR #3031, opened 2024-11-19
Adds Beta Cmdlets for paths under the `directory.publicKeyInfrastructureRoot` tag

timayabi2020:
This PR updates beta version of DirectoryObjects module with new APIs in order to unblock the workload responsible for the below paths who are nearing their GA date.
 ``/directory/publicKeyInfrastructure/certificateBasedAuthConfigurations`` and  /directory/publicKeyInfrastructure/certificateBasedAuthConfigurations/{id}/certificateAuthorities`` 

Below is an image of the module manifest showing the added cmdlets.
image


tools/TweakOpenApi.ps1
```diff
@@ -16,12 +16,12 @@ $prepositionReplacements = @{
     At   = "GraphAPre"
     For  = "GraphFPre"
     Of   = "GraphOPre"
-    Or   = "GraphRPre"
+    #Or   = "GraphRPre" Handling of this preposition needs to be reviewed
```
andrueastman, line 19: Out of curiosity, is there a specific reason to comment this out?

timayabi2020: Yes. I noticed operation ids where ``Or`` doesn't appear as a conjunction was being renamed. For example, an operation id that contains "Org" or "Organization". The review will involve detecting appearance of "Or" as a conjuction only.

andrueastman (approved):

Issue #3069, opened 2025-01-10
Get-MgApplication & Remove-MgApplication inconsistent use of ApplicationId (unable to pipe get to remove)


trevor-harte-ase:
### Describe the bug

The following code does not work:
`Get-MgApplication | ? {$_.displayname -eq 'test'} | Remove-MgApplication`

The response from Remove-MgApplication is InputObject has null value for InputObject.ApplicationId
This response is valid as the output from Get-MgApplication returns Id and AppId, however, I believe that the intention shown by most other Powershell cmdlets is that there should be a consistent capability to pipe one command to another, especially when they are within the same module.

To workarount, I assumed the following would then work:
`Get-MgApplication | ? {$_.displayname -eq 'test'} | % {Remove-MgApplication -ApplicationId $_.AppId}`
This is incorrect as I need to send $_.Id because the API expects ObjectID (as per https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1715 which was closed stating 'by design')

I understand that the API expects ObjectId, so why do these two applications have their parameter name as ApplicationId rather than ObjectId? Also, to assist in piping the commands together, could Remove-MgApplication have an alias for ApplicationId(aka ObjectId) that accepts InputObject.Id?

Finally, can the documentation of the cmdlets be updated so as to remove ambiguity around the usage of Id, ObjectId, ApplicationId and AppId?

Thanks.

### Expected behavior

should be able to pipe output from get-MgApplication to remove-MgApplication

### How to reproduce

Create an application called 'test'
Run connect-MgGraph
Run `Get-MgApplication | ? {$_.displayname -eq 'test'} | Remove-MgApplication`

### SDK Version

2.25.0

### Known Workarounds

Run `Get-MgApplication | ? {$_.displayname -eq 'test'} | % {Remove-MgApplication -ApplicationId $_.Id}`

### Debug output

<details><summary>Click to expand log</summary>
```

<Log output here>

```
</details>

### Configuration

Name                           Value
----                           -----
PSVersion                      7.4.5
PSEdition                      Core
GitCommitId                    7.4.5
OS                             Debian GNU/Linux 12 (bookworm)
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

X64 architecture
[...]

PR #3114, opened 2025-02-10
Adds `-ObjectId` parameter as an alias for `-ApplicationId`

### LINKED ISSUE #2513: Get-MgApplication's -ApplicationId parameter should be called -ObjectId
Attached images explain the problem. 
The following image compares the `Get-MgApplication` to the correct implementation in `Get-AzAdApplication`:
id_or_appid
The second image gives a short analysis of the problem:
final

timayabi2020:
Fixes #2513 
Updates AutoREST directive to add `-ObjectId` parameter as an alias for `-ApplicationId`

image

image


andrueastman (approved):
👍🏼

trevor-harte-ase:
Does this solve the issue of piping the get command to the remove command that I reported in #3069 ?

The issue there was not just the "objectId" alias for -applicationId, but that the output of the get-MgApplication command returns fields "Id" and "AppId" and neither of which was considered valid input for the InputObject.ApplicationId of remove-MgApplication as their names do not match.

I believe solving the issue of piping Get to Remove would require remove-MgApplication to have an alias of "Id" for the existing -ApplicationId

Issue #3008, opened 2024-10-23
null valued properties should be serialized in when using cmdlets


gavinbarron:
### Describe the bug

A customer is trying to clear the Devices Conditions in a Conditional Access Policy and cannot using `Update-MgIdentityConditionalAccessPolicy`

### Expected behavior

We can clear sections of a Conditional Access Policy using `Update-MgIdentityConditionalAccessPolicy`

When running the reproduction code we should see a request body with a null valued Devices property, e.g.:

```
{
  "Conditions": {
    "Devices": null
  }
}
``` 

### How to reproduce

Run the following commands with fiddler set up and intercepting requests to Microsoft Graph
```
Connect-MgGraph
$policyId = "POLICYID"

$params = @{
  conditions = @{
   devices = $null
  }
}

Update-MgIdentityConditionalAccessPolicy -ConditionalAccessPolicyId $policyId -BodyParameter $params
```

You will see that the body of the request sent over the wire contains and empty object `{}`

### Known Workarounds

use `Invoke-GraphRequest`

### Debug output

not provided in IcM: https://portal.microsofticm.com/imp/v5/incidents/details/557480666

### Configuration

Not provided

PR #3126, opened 2025-02-13
Allows serialization of null valued properties.

### LINKED ISSUE #2609: Update-MGUser -agegroup doesn't allow null value despite documentation
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
The Update-MgUser [documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.users/update-mguser?view=graph-powershell-1.0#-agegroup) states that -AgeGroup accepts ``null`` as one of its values. In practice, it does not, erroring out. What does work is ``adult``, ``minor``, and ``notadult``.

**Describe alternatives you've considered**
``update-mguser -userid "somebody@domain.com" -agegroup null``
``update-mguser -userid "somebody@domain.com" -agegroup "null"``
``update-mguser -userid "somebody@domain.com" -agegroup $null``

``Invoke-GraphRequest -Method PATCH -Uri "https://graph.microsoft.com/v1.0/Users/$($myupn)" -Body '{"AgeGroup": []}'``
^This also fails with effectively the same error.

**Additional context**
n/a

timayabi2020:
﻿

Fixes #2609 and #3008 

- Adds autorest directive to ensure that null valued properties are not omitted by default.
- Adds a custom JsonExtensions class containing an extension method that ensures properties explicitly defined as null in a PowerShell session are retained 
- Adds test project to test the above custom class.

This PR has a dependency on the autorest update done here which is part of addressing the issue here https://github.com/microsoftgraph/autorest.powershell/pull/9

image

image

### DIFF
--- a/tools/Tests/JsonUtilitiesTest/JsonExtensionsTests.cs
+++ b/tools/Tests/JsonUtilitiesTest/JsonExtensionsTests.cs
@@ -1,0 +1,128 @@
+﻿namespace JsonUtilitiesTest;
+using System;
+using Newtonsoft.Json.Linq;
+using Xunit;
+using Microsoft.Graph.PowerShell.JsonUtilities;
+
+public class JsonExtensionsTests
+{
+    [Fact]
+    public void RemoveDefaultNullProperties_ShouldRemoveDefaultNullValues()
+    {
+        // Arrange
+        JObject json = JObject.Parse(@"{
+            ""displayname"": ""Tim"",
+            ""position"": ""defaultnull"",
+            ""salary"": 2000000,
+            ""team"": ""defaultnull""
+        }");
+
+        // Act
+        string cleanedJson = json.RemoveDefaultNullProperties();
+        JObject result = JObject.Parse(cleanedJson);
+
+        // Assert
+        Assert.False(result.ContainsKey("position"));
+        Assert.False(result.ContainsKey("team"));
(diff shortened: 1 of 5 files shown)


tools/Tests/JsonUtilitiesTest/JsonExtensionsTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            ""team"": """"
+        }");
+
+        // Act
+        string cleanedJson = json.RemoveDefaultNullProperties();
+        JObject result = JObject.Parse(cleanedJson);
+
+        // Assert
+        Assert.Null(result["position"]?.Value<string>());
+        Assert.Null(result["team"]?.Value<string>());
```
Ndiritu, line 48: Curious whether it's correct to be converting explicitly empty strings to null values in the payload.

timayabi2020: Makes sense. We can remove that conversion for now

Ndiritu: For Kiota SDKs, we retain the explicit empty strings - https://github.com/microsoft/kiota-dotnet/blob/main/src/serialization/json/JsonSerializationWriter.cs#L93

Ndiritu (dismissed):

Ndiritu (approved):

PR #3141, opened 2025-02-18
Updates implementation of allowing null valued properties to handle array and object schema types differently

timayabi2020:
This PR is dependent on the AutoREST PR here https://github.com/microsoftgraph/autorest.powershell/pull/10

It rectifies an error whenever the implementation for allowing null valued properties encounters a body parameter whose schema type is an array. 
**NB**. For the PR gate to pass, the autorest submodule must be updated.


tools/Custom/JsonExtensions.cs
```diff
@@ -1,62 +1,85 @@
...
-                    else if (property.Value.Type == JTokenType.String && property.Value.ToString() == "defaultnull")
-                    {
-                        property.Remove();
-                    }
-                    else if (property.Value.Type == JTokenType.String && (property.Value.ToString() == "null"))
+                }
+                else if (token is JArray jsonArray)
+                {
+                    // Process each item in the JArray
+                    for (int i = jsonArray.Count - 1; i >= 0; i--)
```
andrueastman, line 52: Àny specific reason to iterate this from the end?

tools/Custom/JsonExtensions.cs
```diff
@@ -1,62 +1,85 @@
...
-                            {
-                                RemoveDefaultNullProperties((JObject)item);
-                            }
+                            RemoveDefaultNullProperties(property.Value);
+                        }
+                        else if (property.Value.Type == JTokenType.Array)
+                        {
+                            RemoveDefaultNullProperties(property.Value);
+                        }
+                        else if (property.Value.Type == JTokenType.String && property.Value.ToString() == "defaultnull")
```
andrueastman, line 39: update the string comparisons to pass a comparison type. 
e.g
```suggestion
                        else if (property.Value.Type == JTokenType.String && property.Value.ToString().Equeals("defaultnull",StringComparison.Ordinal)
```

tools/Custom/JsonExtensions.cs
timayabi2020: It's just a personal practice to ensure that the property is deleted safely. i.e, to prevent any possibility of skipping.

andrueastman (approved):

PR #3171, opened 2025-02-24
Roll back to .NET 6

timayabi2020:
This PR rolls back the authentication module to .NET 6 due to the breaking change in the .NET SDK version 8.x. which is not compatible with Azure Automation Run Book's PowerShell 7.2 version.

### DIFF
--- a/.azure-pipelines/common-templates/install-tools.yml
+++ b/.azure-pipelines/common-templates/install-tools.yml
@@ -5,7 +5,12 @@ steps:
   - task: UseDotNet@2
     displayName: Use .NET SDK
     inputs:
-      version: 6.x     
+      version: 8.x 
+
+  - task: UseDotNet@2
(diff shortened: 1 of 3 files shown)


baywet (approved):

nkasco:
This type of change, while warranted, really needs to be better documented as a breaking change so that users have full awareness without digging.

So am I interpreting correctly that because another MS product is not up to date, that this change was rolled back and the SDK will continue to work on 7.2 and 7.3?

o-l-a-v:
Actually, Automation Account supports PowerShell 7.4, has done since november 2024. It's just not documented anywhere, and thus probably not GA?

* <https://github.com/PowerShell/PowerShell/discussions/24464#discussioncomment-11268201>
* <https://github.com/PowerShell/PowerShell/discussions/25000#discussioncomment-12267372>

PR #3330, opened 2025-05-30
Unblock PowerShell Generation in the Calendar Module

ramsessanchez:
Adding two regex expressions to the Directives list which will remove paths from generation based on Operation Id matches: 

.*exceptionOccurrence.*$|
*\.instance(_|\.).*$|

1. Anything with `exceptionOccurrence` in the operation id, this is because `exceptionOccurrence` shouldn't exist in the URI path at all as it is a navigation property. 

2. Anything where the operation id includes `path .instance.` or `.instance_`. This is because instance shouldn't be followed by path segment in the URI, instead it should always be followed by a datetime window query according to documentation. `instance_` implies that the URI has the following `instance/{event-id}/` which we know isn't allowed since instance needs to be at the end and followed by a date time window query.


MIchaelMainer:
To be clear....
1. The reason why exceptionOccurrences is not part of an id is that it isn't used as a segment. That navigation is only accessed [via select and expand](https://learn.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0#relationships).

PR #3385, opened 2025-08-20
Update example cmdlet in New-MgDomainFederationConfiguration

juntakata:
To properly configure a federation, PreferredAuthenticationProtocol option is required. This is explained in [Create internalDomainFederation](https://learn.microsoft.com/en-us/graph/api/domain-post-federationconfiguration?view=graph-rest-1.0&tabs=http) as below. As the sample cmdlet is missing this parameter. I added it to the example.

image

Also, most customers want to handle MFA at their IdP. The example has -FederatedIdpMfaBehavior "rejectMfaByFederatedIdp" set that means MFA done in their IdP is blocked by Entra ID. I would like to change it to "acceptIfMfaDoneByFederatedIdp" to cover more popular user scenarios.

### DIFF
--- a/src/Identity.DirectoryManagement/v1.0/examples/New-MgDomainFederationConfiguration.md
+++ b/src/Identity.DirectoryManagement/v1.0/examples/New-MgDomainFederationConfiguration.md
@@ -1,7 +1,7 @@
 ### Example 1: Configure federation settings for a federated domain
 
 ```powershell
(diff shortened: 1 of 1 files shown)


juntakata:
@MIchaelMainer I am not sure if you are the right person, but can you review this doc PR? Multiple customers are in trouble due to the missing option. This update will help customers.

PR #3407, opened 2025-09-23
Fix: Include the users onPremSyncBehavior API in the OpenAPI spec

darshanmaiya:
The new /users/{id}/onPremisesSyncBehavior path is not being included in the Beta.Users module. This is due to the way the Open API specs are fetched for generation.

This PR adds the tag for that operation into the set of tags fetched.


darshanmaiya:
@ramsessanchez : Could you please help review this PR?

PR #3430, opened 2025-10-22
Split Weekly Generation Pipeline

ramsessanchez:
Split Weekly generation pipeline into two separate pipelines, one to retrieve openApi updates and individually build the modules, and a second pipeline to generate the comandMetadata which requires all modules build on a single image. 

The following yml files handle the following: 
1.) downloading the most recent openApi docs 
2.) determine which modules have been updated
3.) build the updated modules independently to validate openApi document correctness
4.) if all updated modules build successfully then a PR is created with the updated openApi documents. 
`Weekly-Generation.yml` 
`Individualized-Workload-Modules.yml`

The second will be triggered on merges to Dev and can also be triggered manually. 
1.) All modules are built on a single image 
2.) the commandMetadata is generated after all modules are built. 
3.) successful completion of both results in a pull request into the Dev branch with an updated commandMetadata file. 
`Module-Metadata-generation.yml`
 
Note: The second pipeline is still in testing as we can't test the pipeline fully until we have it registered as one of our pipelines in devops and add the appropriate token.

### DIFF
--- a/config/ModulesMapping.jsonc
+++ b/config/ModulesMapping.jsonc
@@ -26,7 +26,7 @@
   "Identity.Partner": "^tenantRelationships.delegatedAdminRelationship$|^tenantRelationships.delegatedAdminCustomer$",
   "Mail": "^users.inferenceClassification$|^users.mailFolder$|^users.message$",
   "ManagedTenants": "^tenantRelationships.managedTenant$",
-  "Migrations": "^solutions.migrations",
+  "Migrations": "^solutions.migrationsRoot",
   "NetworkAccess": "^networkAccess\\.",
   "Notes": "^users.onenote$|^groups.onenote$|^sites.onenote$",
   "People": "^users.person$|^users.profile$|^users.officeGraphInsights$|^users.userAnalytics$",


ramsessanchez:
The following demonstrates ability to build modules individually: 
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=200425&view=results
The following demonstrates ability to specify which modules to build and create pr: 
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=200930&view=results
The following pr was made by the previous pipeline run, no chages to the openApi docs as updating the openApi doc step was intentionally skipped. https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/3429

MIchaelMainer:
@ramsessanchez do we have any preliminary pipeline runs that use these updated scripts?

ramsessanchez:
> @ramsessanchez do we have any preliminary pipeline runs that use these updated scripts?

Assuming you're talking about the script in download-openApi .yml? No other pipelines use this outside of the weekly generation pipeline. The update focuses on making sure that the list of modules in the diff are listed with v1.0 or beta label, for the purpose of the module build matrix.

.azure-pipelines/common-templates/download-openapi-docs.yml
```diff
@@ -73,12 +73,16 @@ steps:
...
           $diff | %{
-            if (($_ -match 'openApiDocs\/(v1.0|beta)\/(.*).yml') -and !$ModulesWithChanges.ContainsKey($matches.2))
-            {
-              $ModulesWithChanges.Add($matches.2, $matches.1)
+            if ($_ -match 'openApiDocs\/(v1.0|beta)\/(.*)\.yml') {
+              $version = if ($matches[1] -eq 'v1.0') { 'v1.0' } else { 'beta' }
+              $moduleName = "$($matches[2])_$version"
+              if (!$ModulesWithChanges.ContainsKey($moduleName)) {
+                $ModulesWithChanges.Add($moduleName, $matches[1])
+              }
```
gavinbarron: Seeing a script this complex inline in a pipeline makes me a bit nervous.
yaml is not a great authoring tool for multiline scripts. 
If this were refactored into a stand-alone script that you could call then your can test it independent of a pipeline run.
Plus you could eliminate the if statment in here an use a condition on the step combined with setting the ModulesWithChanges variable to 'Skipped' by default.

ramsessanchez: Gotcha. Included the above runs as references showing that the scripts did work. Initially I was just editing the inline scripts that were already there, however as the complexity grew they definitely should have been moved to their own file. Will update with the changes.

PR #3467, opened 2025-12-12
Separate weekly openApi-doc refresh and command-metadata refresh

ramsessanchez:
Split Weekly generation pipeline into two separate pipelines, one to retrieve openApi updates and individually build the modules, and a second pipeline to generate the comandMetadata which requires all modules build on a single image.

The following yml files handle the following:
1.) downloading the most recent openApi docs
2.) determine which modules have been updated
3.) build the updated modules independently to validate openApi document correctness
4.) if all updated modules build successfully then a PR is created with the updated openApi documents.
Weekly-Generation.yml
Individual-Workload-Module.yml

The second will be triggered on merges to Dev and can also be triggered manually.
1.) All modules are built on a single image
2.) the commandMetadata is generated after all modules are built.
3.) successful completion of both results in a pull request into the Dev branch with an updated commandMetadata file.
command-metadata-refresh.yml

Note: The second pipeline is still in testing as we can't test the pipeline fully until we have it registered as one of our pipelines in devops and add the appropriate token.

### DIFF
--- a/.github/policies/msgraph-sdk-powershell-branch-protection.yml
+++ b/.github/policies/msgraph-sdk-powershell-branch-protection.yml
@@ -9,41 +9,6 @@ resource: repository
 configuration:
   branchProtectionRules:
 
-  - branchNamePattern: dev
-    # This branch pattern applies to the following branches as of approximately 08/25/2023 14:07:25:
-    # dev
-
-    # Specifies whether this branch can be deleted. boolean
-    allowsDeletions: false
-    # Specifies whether forced pushes are allowed on this branch. boolean
-    allowsForcePushes: false
-    # Specifies whether new commits pushed to the matching branches dismiss pull request review approvals. boolean
-    dismissStaleReviews: true
-    # Specifies whether admins can overwrite branch protection. boolean
-    isAdminEnforced: false
-    # Indicates whether "Require a pull request before merging" is enabled. boolean
-    requiresPullRequestBeforeMerging: true
-    # Specifies the number of pull request reviews before merging. int (0-6). Should be null/empty if PRs are not required
-    requiredApprovingReviewsCount: 1
-    # Require review from Code Owners. Requires requiredApprovingReviewsCount. boolean
-    requireCodeOwnersReview: true
-    # Are commits required to be signed. boolean. TODO: all contributors must have commit signing on local machines.
-    requiresCommitSignatures: false
-    # Are conversations required to be resolved before merging? boolean
-    requiresConversationResolution: true
-    # Are merge commits prohibited from being pushed to this branch. boolean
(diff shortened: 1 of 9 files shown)


.azure-pipelines/common-templates/download-openapi-docs.yml
```diff
@@ -23,74 +19,46 @@ steps:
...
+        # Set the Branch pipeline variable
+        Write-Host "##vso[task.setvariable variable=Branch;isOutput=false]$branch"
 
   - task: Bash@3
     displayName: "Create weekly branch"
     inputs:
       targetType: inline
       script: |
         git status
-        git branch $(ComputeBranch.WeeklyBranch)
```
MIchaelMainer: Is the task variable name optional for specifying an environment variable set from an ADO task?  I haven't seen this before

.azure-pipelines/command-metadata-refresh.yml
```diff
@@ -0,0 +1,150 @@
...
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool: $(BuildAgent)
+    sdl:
+      binskim:
+        enabled: false
+        justificationForDisabling: "Binskim keeps on crashing and failing the weekly build pipeline. Disabling it for now because we are unable to publish the artifacts to internal feeds."
```
MIchaelMainer: Do we understand why it is failing?

.azure-pipelines/common-templates/download-openapi-docs.yml
```diff
@@ -4,15 +4,11 @@
 parameters:
   - name: Branch
     type: string
+    default: "WeeklyApiRefresh"
   - name: BaseBranch
     type: string
   - name: BuildAgent
     displayName: Build Agent
-  - name: SkipForceRefresh
```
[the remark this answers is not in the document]
ramsessanchez: Further, this pipeline is called by the weeklyGeneration pipeline. That pipeline contains the skipDownload parameter, which will skip this workflow altogether if set to True. Therefore, these parameters are unnecessary here and consolidated in the weeklyGeneration pipeline.

ramsessanchez:
> I'm curious how much this reduces the average generation run.

This won't necessarily reduce generation time. Instead, this will ensure that we are able to catch instances of failure in the modules individually. At the moment, during cases where multiple modules fail, we are only alerted of the first failure. Because all others are hidden, we can only know about one failure at a time bottlenecking us to tackling one failure at a time. This was we can know all the failures at once and tackle them in parallel.

MIchaelMainer (approved):

PR #3472, opened 2025-12-19
Feat/force wam

ramsessanchez:
Enabling WAM on windows systems by default. This cannot be disabled on Windows systems. Using `Set-MgGraphOption` with the value `$False` will have no impact on Windows systems.

### DIFF
--- a/src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
+++ b/src/Authentication/Authentication/Cmdlets/ConnectMgGraph.cs
@@ -275,6 +275,7 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
             stringBuilder.AppendLine($"SDK Docs: {Constants.SdkDocsLink}");
             stringBuilder.AppendLine($"API Docs: {Constants.ApiDocsLink}{System.Environment.NewLine}");
             stringBuilder.AppendLine($"NOTE: You can use the -NoWelcome parameter to suppress this message.");
+            stringBuilder.AppendLine($"NOTE: Sign in by Web Account Manager (WAM) is enabled by default on Windows systems and cannot be disabled. Any setting stating otherwise will be ignored.");
             return stringBuilder.ToString();
         }
 
@@ -342,7 +343,6 @@ namespace Microsoft.Graph.PowerShell.Authentication.Cmdlets
         {
             GraphSessionInitializer.InitializeSession();
             GraphSession.Instance.DataStore = new DiskDataStore();
-            GraphSession.Instance.OutputWriter.WriteWarning("Note: Sign in by Web Account Manager (WAM) is enabled by default on Windows systems and cannot be disabled. Any setting stating otherwise will be ignored.");
         }
 
         /// <summary>


src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -129,16 +129,13 @@ private static async Task<InteractiveBrowserCredential> GetInteractiveBrowserCre
                     authRecord = await Task.Run(() =>
                     {
                         // Run the thread in MTA.
+                        GraphSession.Instance.OutputWriter.WriteWarning("Note: Signin by Web Account Manager (WAM) is enabled by default on Windows. If using an embedded PowerShell terminal, the interactive browser window may be hidden behind other windows.");
                         return interactiveBrowserCredential.Authenticate(new TokenRequestContext(authContext.Scopes), cancellationToken);
                     });
                 }
                 else
                 {
-                    authRecord = await Task.Run(() =>
```
FehintolaObafemi: Why was the MTA run removed?

gavinbarron: A better question is why was an async call wrapped in a Task.Run in the first place? 

The `AuthenticateAsync` method being called here returns `Task<AuthenticationRecord>` where as the `Authenticate` method above is synchronous and needs to wrapped to allow it to run on a separate thread and be awaited.

src/Authentication/Authentication/Cmdlets/SetMgGraphOption.cs
```diff
@@ -25,7 +25,8 @@ protected override void ProcessRecord()
             if (this.IsParameterBound(nameof(EnableLoginByWAM)))
             {
                 GraphSession.Instance.GraphOption.EnableWAMForMSGraph = EnableLoginByWAM;
-                WriteDebug($"Signin by Web Account Manager (WAM) is {(EnableLoginByWAM ? "enabled" : "disabled")}.");
+                WriteDebug($"Requested to {(EnableLoginByWAM ? "enable" : "disable")} signin by Web Account Manager (WAM).");
+                WriteDebug("Note: Signin by Web Account Manager (WAM) is mandatory for Microsoft Graph PowerShell SDK on Windows and cannot be disabled. The option will be set to enabled by default on Windows.");
```
gavinbarron:
```suggestion
                WriteDebug("Note: Sign in by Web Account Manager (WAM) is mandatory for Microsoft Graph PowerShell SDK on Windows and cannot be disabled. This option is deprecated and no longer has any effect.");
```

gavinbarron (changes requested):

src/Authentication/Authentication/Cmdlets/SetMgGraphOption.cs
ramsessanchez: Is it deprecated if still functional on other systems?

src/Authentication/Authentication/Cmdlets/SetMgGraphOption.cs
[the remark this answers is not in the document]
ramsessanchez: done

gavinbarron (approved):

nkasco:
Please see #3481 caused by this PR

PrzemyslawKlys:
Please revert this PR, it breaks more then it solves.

PR #3504, opened 2026-01-22
fix: catch error codes in parallel

ramsessanchez:
Catch all generation and build failures in the scripts and report them at the end. Rather than stopping execution right away, catch failures at the end and report them. 
These scripts are used in our CI and Weekly generation pipelines. This change aims to more clearly report failed modules and ensure all failures are caught on a single run.


ramsessanchez:
@gavinbarron unfortunately, after testing, this does not catch IOExceptions and some other exceptions rooted in autorest, moving this to draft again in the meantime.

ramsessanchez:
@gavinbarron , will tackle IO exception issues in separate PR.

PR #3505, opened 2026-01-22
Feature: Add ability to disable WAM 

ramsessanchez:
WAM can now be disabled in cases where the Client Id being used is not the default value.

### DIFF
--- a/src/Authentication/Authentication/Models/AuthContext.cs
+++ b/src/Authentication/Authentication/Models/AuthContext.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Graph.PowerShell.Authentication
         public SecureString ClientSecret { get; set; }
         public string Environment { get; set; } = GraphEnvironmentConstants.EnvironmentName.Global;
         private bool IsCustomClientId => !string.Equals(ClientId, PowerShellClientId, StringComparison.OrdinalIgnoreCase);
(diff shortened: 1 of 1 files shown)


src/Authentication/Authentication/Cmdlets/SetMgGraphOption.cs
gavinbarron: This implementation feels lacking. 

Because `EnableLoginByWAM` is a bool it defaults to false. So simply not setting this gets you to bypass WAM which is not the intent. 

The user MUST explicitly use `Set-MgOption -EnableLoginByWAM $false` in order to bypass WAM with their own appId

src/Authentication/Authentication/Cmdlets/SetMgGraphOption.cs
```diff
@@ -26,7 +26,10 @@ protected override void ProcessRecord()
             {
                 GraphSession.Instance.GraphOption.EnableWAMForMSGraph = EnableLoginByWAM;
                 WriteDebug($"Requested to {(EnableLoginByWAM ? "enable" : "disable")} sign in by Web Account Manager (WAM).");
-                WriteDebug("Note: Sign in by Web Account Manager (WAM) is mandatory for Microsoft Graph PowerShell SDK on Windows and cannot be disabled. This option is deprecated and no longer has any effect.");
+                if (!EnableLoginByWAM)
```
gavinbarron:
```suggestion
                if (EnableLoginByWAM.HasValue && !EnableLoginByWAM.Value)
```

PR #3548, opened 2026-03-05
fix: update msal dependencies

gavinbarron:
﻿Updates MSAL dependencies to latest to ensure that non-broker interactive browser based auth flow uses form_post to send auth code to the client rather than a redirect which exposes the auth code in the URL.


ramsessanchez:
looks good, but lets hold off on merging until issues with the CI pipeline are resolved

PR #3584, opened 2026-04-21
Ramsess/update beta open api manual 04212026

ramsessanchez:
remove psd1 file tracking, these are not needed as they are generated during build and overwritten in pipelines. 
manually update DeviceManagement.Enrollment beta open api doc
manually update WindowsUpdates beta open api doc and add directives to fix generation issues


ramsessanchez:
Superseded by #3601

PR #3593, opened 2026-04-24
Ramsess/fix release pipeline

ramsessanchez:
Ensure all npm calls are skipped during autorest generation release by pre-downloading dependencies from private registry so that the pipeline is compliant with network isolation policies in release job workflow.

### DIFF
--- a/.azure-pipelines/common-templates/install-tools.yml
+++ b/.azure-pipelines/common-templates/install-tools.yml
@@ -52,18 +52,21 @@ steps:
     displayName: Install AutoRest
     inputs:
       command: custom
+      workingDir: $(Build.SourcesDirectory)
       customCommand: install -g autorest@3.7.2
-    
+
   - task: Npm@1
     displayName: Install AutorestCore
     inputs:
       command: custom
+      workingDir: $(Build.SourcesDirectory)
       customCommand: install -g @autorest/core@3.10.4
 
   - task: Npm@1
     displayName: Install Rush
     inputs:
       command: custom
+      workingDir: $(Build.SourcesDirectory)
       customCommand: install -g @microsoft/rush
   
   - task: PowerShell@2
--- a/.azure-pipelines/common-templates/install-tools.yml
+++ b/.azure-pipelines/common-templates/install-tools.yml
@@ -66,14 +66,14 @@ steps:
     inputs:
       command: custom
       workingDir: $(Build.SourcesDirectory)
-      customCommand: install -g autorest@3.7.2 --registry https://microsoftgraph.pkgs.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_packaging/PowerShell_V2_Build/npm/registry/
+      customCommand: install -g autorest@3.7.2
 
   - task: Npm@1
     displayName: Install AutorestCore
     inputs:
       command: custom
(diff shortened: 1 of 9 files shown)


ramsessanchez:
validating via: https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=216781&view=results

.azure-pipelines/common-templates/install-tools.yml
```diff
@@ -48,23 +48,70 @@ steps:
...
 
+  - task: PowerShell@2
+    displayName: Apply npm auth config to user profile
+    inputs:
+      targetType: inline
+      pwsh: true
+      script: |
+        # Copy authenticated .npmrc to the user home dir so ALL npm subprocesses
+        # (including autorest's internal npm calls) use the private feed + auth tokens.
+        $src = "$(Build.SourcesDirectory)/.npmrc"
```
gavinbarron, line 59: where is this authenticat3ed .npmrc file? I don't see it as an artifact anywhere in the repo

.azure-pipelines/common-templates/install-tools.yml
```diff
@@ -48,23 +48,70 @@ steps:
...
+        Write-Host "Copied npm config to $dst"
+
   - task: Npm@1
     displayName: Install AutoRest
     inputs:
       command: custom
-      customCommand: install -g autorest@3.7.2
-    
+      workingDir: $(Build.SourcesDirectory)
+      customCommand: install -g autorest@3.7.2 --registry https://microsoftgraph.pkgs.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_packaging/PowerShell_V2_Build/npm/registry/
```
gavinbarron: shouldn't this just point at the authenticated .npmrc file and thus avoid having this same URL duplicated all over the place?

.azure-pipelines/common-templates/install-tools.yml
gavinbarron: Answering my own question here, derp, I added a step to the pipeline that creates it dynamically during pipeline execution.

.azure-pipelines/common-templates/install-tools.yml
ramsessanchez: this was required due to the -global install, since global install wouldn't use the local authentication file. Changed to a local install and this flag is no longer needed.

.azure-pipelines/common-templates/install-tools.yml
```diff
@@ -48,23 +48,70 @@ steps:
...
+      workingDir: $(Build.SourcesDirectory)
+      customCommand: install autorest@3.7.2
+
   - task: Npm@1
     displayName: Install AutorestCore
     inputs:
       command: custom
-      customCommand: install -g @autorest/core@3.10.4
+      workingDir: $(Build.SourcesDirectory)
+      customCommand: install @autorest/core@3.10.4
```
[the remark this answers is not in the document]
gavinbarron: @ramsessanchez looks like there's a good reason not to use local installs over global installs, that said I think you can use the -c option to pass the .npmrc file that points at the private registry

gavinbarron (approved):
LGTM

PR #3603, opened 2026-05-08
bump node install tool and node version

ramsessanchez:
Node@0 is deprecated and needs to be incremented to v1
Node 18 has reached EOL and needs to be bumped to LTS v22. 
Removing Rush node restriction so node22 can be used https://github.com/microsoftgraph/autorest.powershell/pull/17


gavinbarron (approved):

ramsessanchez:
will resolve in follow up pr

Issue #3394, opened 2025-09-10
Connect-MgGraph fails after Connect-ExchangeOnline


OtterKring:
### Describe the bug

I don't know, if this is related to the ExchangeOnlineManagement module (my suspicison) or to Microsoft.Graph, but you should know; too, I guess:

I updated Microsoft.Graph from 2.29.1. to 2.30.0 yesterday and ExchangeOnlineManagement from 3.8.0 to 3.9.0. After the update I found that when I run Connect-ExchangeOnline before Connect-MgGraph, the latter fails with the following error:
```
Connect-MgGraph: ClientCertificateCredential authentication failed: Method not found: '!0 Microsoft.Identity.Client.BaseAbstractApplicationBuilder`1.WithLogging(Microsoft.IdentityModel.Abstractions.IIdentityLogger, Boolean)'.
```

When I run Connect-MgGraph first and then Connect-ExchangeOnline, it works.

### Expected behavior

IMO it should not matter, which command is run first.

### How to reproduce

- On Powershell 7.5.2 ... 
- Install Microsoft.Graph 2.30.0
- Install ExchangeOnlineManagement 3.9.0
- First run Connect-ExchangeOnline. ( I used certificate authentication )
- Second run Connect-MgGraph. It does not matter if user or certificate authentication. I tried both
- You should get the error described above

### SDK Version

2.30.0

### Latest version known to work for scenario above?

2.29.1 together with ExchangeOnlineManagement 3.8.0

### Known Workarounds

1) First run Connect-MgGraph
2) Then run Connect-ExchangeOnline

### Debug output

<details><summary>Click to expand log</summary>
```

DEBUG: ClientCertificateCredential.GetToken invoked. Scopes: [ https://graph.microsoft.com/.default ] ParentRequestId:
DEBUG: ClientCertificateCredential.GetToken was unable to retrieve an access token. Scopes: [ https://graph.microsoft.com/.default ] ParentRequestId:  Exception: Azure.Identity.AuthenticationFailedException (0x80131500): ClientCertificateCredential authentication failed: Method not found: '!0 Microsoft.Identity.Client.BaseAbstractApplicationBuilder`1.WithLogging(Microsoft.IdentityModel.Abstractions.IIdentityLogger, Boolean)'.
[...]

Issue #2978, opened 2024-09-30
Implement proper assembly isolation to play nice with `vscode-powershell`


o-l-a-v:
### Is your feature request related to a problem? Please describe the problem.

According to `vscode-powershell` maintainers, Azure PowerShell (and I assume `Microsoft.Graph` too as I see similar problems with it) does not "implement assembly isolation correctly across the board".

* <https://github.com/PowerShell/vscode-powershell/issues/4727#issuecomment-1762104091>

This can cause assembly conflict issues that `vscode-powershell` can't solve easily from their context.

I don't have deep knowledge about this, I just want my authoring experience where Microsoft owns every part of the stack fixed.

Here is more context:

* 2020-10-15: <https://github.com/PowerShell/vscode-powershell/issues/3012>
* 2023-09-05: <https://github.com/PowerShell/vscode-powershell/issues/4727>
* 2024-09-30: <https://github.com/PowerShell/vscode-powershell/issues/5053>

### Describe the solution you'd like.

Find out if anything can be done here to make Microsoft.Graph modules play nice with `vscode-powershell`.

RobSEIT:
Holy shit, yes please fix this. I've been tearing my hair out for months due to conflicts with 365 modules like Graph, Az Accounts (which is different to the Azure AD module that is being replaced by Graph), Exchange Online, Partner Center (even before it was officially deprecated)

The only fixes I've found are to either bash my head against the wall updating, uninstalling/reinstalling modules, trying random versions or killing and restarting the PowerShell extension in VScode. Or making separate scripts and running them in PowerShell 5 from the built in ISE, then saving the output and loading in another script in PowerShell 7 in vscode

SamErde:
I just released v1.0 of my [DLL Pickle](https://github.com/SamErde/DLLPickle) module that prevents most cases of these DLL version conflicts. Please try it out and let me know your results!

```powershell
Install-Module -Name 'DLLPickle' -Scope CurrentUser
Import-Module -Name 'DLLPickle'
Import-DPLibrary
```

PrzemyslawKlys:
While we wait for MS to fix it I've added ALC automated handling to my PowerShell builder `Install-Module PSPublishModule`

```powershell
Import-Module PSPublishModule -Force

$exo = Import-IsolatedModule -Profile ExchangeOnlineManagement -PassThru
$teams = Import-IsolatedModule -Profile MicrosoftTeams -PassThru
[...]

PR #3632, opened 2026-06-03
Isolate Microsoft.Graph.Authentication assembly loading

PrzemyslawKlys:

- Load `Microsoft.Graph.Authentication.dll` through a named `AssemblyLoadContext` on PowerShell Core before importing it as a binary module.
- Register a small managed dependency resolver for that load context so async authentication paths can resolve packaged dependencies from worker threads without requiring a PowerShell runspace.
- Keep Windows PowerShell/Desktop behavior unchanged by continuing to import the binary module by path when `AssemblyLoadContext` is unavailable.
- Preserve the public command and alias exports from the script module wrapper without re-importing the binary module by path.
- Stop adding the authentication assemblies as generated `NestedModules`, because that causes PowerShell to load the DLLs before the script module can choose the load context.
- Add focused module import tests that verify the root authentication assembly and worker-thread dependency resolution stay outside `AssemblyLoadContext.Default` on PowerShell Core.

### Why

This is intended as a narrow assembly-isolation improvement for Microsoft 365 module interop scenarios where another module has already loaded incompatible authentication dependencies in the same PowerShell process. The existing module initializer resolver can help with dependency resolution after the root binary is loaded, but it cannot prevent the root authentication assembly itself from being loaded into the default context when the manifest/import path loads the DLL first.

By letting the script module load the root binary into a custom context first, the authentication module has a chance to keep its dependency graph isolated while preserving the existing exported cmdlet surface.

### Validation

- `pwsh -NoProfile -File .\src\Authentication\Authentication\build-module.ps1 -Release`
- `pwsh -NoProfile -Command 'Invoke-Pester .\src\Authentication\Authentication\test\Microsoft.Graph.Authentication.Tests.ps1 -Output Detailed'`
  - 8 tests passed, including dependency resolution from a worker thread into the isolated load context.
- Artifact import smoke test confirmed:
  - `Microsoft.Graph.Authentication` was not present in `AssemblyLoadContext.Default`
  - the active context name was `Microsoft.Graph.Authentication.<hash>`
  - `Connect-MgGraph` exported successfully
  - `Connect-Graph` and `Invoke-MgRestMethod` aliases resolved correctly
- Live Graph auth smoke on Windows PowerShell 7.6.2 with WAM:
  - `Connect-MgGraph -Scopes User.Read -NoWelcome` succeeded
  - `Azure.Identity`, `Azure.Identity.Broker`, `Microsoft.Identity.Client`, `Microsoft.Identity.Client.Broker`, and `Microsoft.Identity.Client.Extensions.Msal` loaded in the Graph authentication ALC rather than default context.
- Live Exchange-first interop smoke:
  - `Import-Module ExchangeOnlineManagement; Connect-ExchangeOnline` loaded Exchange MSAL/Broker assemblies in default context.
  - Importing the PR-built `Microsoft.Graph.Authentication` artifact and running `Connect-MgGraph -Scopes User.Read -NoWelcome` then succeeded.
  - Graph loaded its own MSAL/Broker/Azure.Identity assemblies in the Graph authentication ALC while Exchange assemblies remained in default context.
- `git diff --check`

### Other links

- Related to #2978
- Related to #3394
- Related to #3576
- Internal validation PR: https://github.com/PrzemyslawKlys/msgraph-sdk-powershell/pull/1

### DIFF
--- a/src/Authentication/Authentication/Microsoft.Graph.Authentication.psm1
+++ b/src/Authentication/Authentication/Microsoft.Graph.Authentication.psm1
@@ -1,7 +1,163 @@
 
-# Load the module dll
+function Get-GraphAuthenticationLoadContextName {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string] $ModulePath
+    )
+
+    $sha256 = [System.Security.Cryptography.SHA256]::Create()
+    try {
+        $pathBytes = [System.Text.Encoding]::UTF8.GetBytes($ModulePath)
+        $hash = [System.BitConverter]::ToString($sha256.ComputeHash($pathBytes)).Replace('-', '').Substring(0, 16)
+        "Microsoft.Graph.Authentication.$hash"
+    }
+    finally {
+        $sha256.Dispose()
+    }
+}
+
+function Initialize-GraphAuthenticationAssemblyResolver {
+    if ('Microsoft.Graph.PowerShell.Authentication.Loader.GraphAuthenticationAssemblyResolver' -as [type]) {
+        return
+    }
+
+    Add-Type -TypeDefinition @'
+using System;
+using System.Collections.Concurrent;
+using System.IO;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace Microsoft.Graph.PowerShell.Authentication.Loader
+{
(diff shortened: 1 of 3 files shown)


ramsessanchez:
merging to different branch to kick off tests in ADO

PR #3633, opened 2026-06-04
Isolate Graph authentication assembly loading (#3632)

ramsessanchez:
Full Description available here: https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/3632


PrzemyslawKlys:
Opened follow-up PR #3635 against this branch to address the Copilot inline comments here and the earlier worker-thread test flakiness note from #3632. Focused auth Pester tests pass locally (8/8) and git diff --check is clean.

PR #3649, opened 2026-06-22
fix: Clear persisted MSAL token cache on Disconnect-MgGraph

### LINKED ISSUE #3648: Disconnect-MgGraph does not clear the persisted MSAL token cache
## Description

When a user runs `Disconnect-MgGraph`, the MSAL token cache persisted to disk is not cleared. Currently, `LogoutAsync()` only:

1. Clears the in-memory token cache
2. Nulls out the AuthContext and GraphHttpClient
3. Deletes the authentication record file (`mg.authrecord.json`)

The disk-persisted MSAL cache (used when `ContextScope == CurrentUser`) at `%LOCALAPPDATA%\.IdentityService\mg.msal.cache.cae` and `mg.msal.cache.nocae` remains intact. This means cached tokens survive disconnection and can be reused in subsequent sessions without re-authentication.

## Expected Behavior

Running `Disconnect-MgGraph` should fully clear all cached authentication state, including the persisted MSAL token cache on disk, so that subsequent `Connect-MgGraph` calls require fresh authentication.

## Current Behavior

After `Disconnect-MgGraph`, the persisted token cache files remain on disk. A new PowerShell session can silently reuse cached tokens without prompting for authentication.

## Suggested Approach

Azure.Identity does not expose APIs to clear the persisted token cache. However, the MSAL library (`Microsoft.Identity.Client.Extensions.Msal`) that Azure.Identity depends on can be used directly:

1. Create `StorageCreationProperties` matching Azure.Identity's internal cache configuration
2. Build a `MsalCacheHelper` targeting the named cache
3. Call `MsalCacheHelper.Clear()` to wipe both `.cae` and `.nocae` cache variants during disconnect

## Environment

- Module: Microsoft.Graph.Authentication
- Affected scope: `ContextScope.CurrentUser` (disk-persisted cache)
- Platforms: Windows (DPAPI), macOS (Keychain), Linux (libsecret/plaintext)

gavinbarron:
## Summary

Fixes #3648

When `Disconnect-MgGraph` is called with `ContextScope == CurrentUser`, the persisted MSAL token cache on disk was not being cleared. This PR adds cache clearing using `MsalCacheHelper.Clear()` from the MSAL Extensions library.

## Changes

- **Added** explicit `Microsoft.Identity.Client.Extensions.Msal` v4.78.0 package reference to `Authentication.Core.csproj`
- **Added** `TokenCacheUtilities.cs` — builds `StorageCreationProperties` matching Azure.Identity's internal cache layout (Windows DPAPI, macOS Keychain, Linux KeyRing) and clears both `.cae` and `.nocae` cache variants
- **Modified** `AuthenticationHelpers.LogoutAsync()` — calls `TokenCacheUtilities.ClearPersistedTokenCacheAsync()` when `ContextScope == CurrentUser`, with non-fatal error handling (disconnect still succeeds if cache clear fails)
- **Added** unit tests for the new behavior
- **Updated** `Disconnect-MgGraph` documentation

## Design Decisions

- **Non-fatal error handling**: Cache clearing failure does not prevent disconnect from succeeding — it's wrapped in try/catch
- **Both CAE variants cleared**: Azure.Identity creates separate `.cae` and `.nocae` cache files; both are cleared
- **Matches Azure.Identity internals**: `StorageCreationProperties` are configured to match the exact storage layout Azure.Identity uses (directory, keychain service, keyring schema/attributes)

## Testing

- 3 new unit tests added (all pass on net8.0 and net472)
- All existing 77 tests continue to pass on net8.0


jsquire (commented):
This looks correct to me, though I don't have a deep understanding in the area.   I've asked @g2vinay and @JonathanCrd to also review.   I'd consider them to be the authoritative voices from the `Azure.Identity` perspective.

ramsessanchez (dismissed):

ramsessanchez:
Approving the changes, however, in line with @jsquire 's comments, we should await feedback from the folks he mentioned. In order to reduce the change of needing a patch release later on, lets await their sign-off and before merging and releasing.

gavinbarron:
Agreed

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -430,6 +430,18 @@ public static async Task<IAuthContext> LogoutAsync()
         {
             var authContext = GraphSession.Instance.AuthContext;
             GraphSession.Instance.InMemoryTokenCache?.ClearCache();
+            if (authContext?.ContextScope == ContextScope.CurrentUser)
+            {
+                try
+                {
+                    await TokenCacheUtilities.ClearPersistedTokenCacheAsync(Constants.CacheName).ConfigureAwait(false);
```
JonathanCrd: on Windows, WAM is on by default for interactive sign-in, and WAM keeps the account + refresh token in its own cache. So cleaning this file won't actually sign the user out of WAM.

For the broker we would need to ask WAM to forget the account. Is the Windows Broker in scope for this fix?

Copilot suggested something like:
```csharp
if (ShouldUseWam(authContext))
{
    var pca = PublicClientApplicationBuilder
        .Create(authContext.ClientId)
        .WithAuthority(GetAuthorityUrl(authContext))
        .WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows))
        .Build();
    foreach (var account in await pca.GetAccountsAsync().ConfigureAwait(false))
        await pca.RemoveAsync(account).ConfigureAwait(false);
}
```

src/Authentication/Authentication.Core/Utilities/TokenCacheUtilities.cs
```diff
@@ -0,0 +1,67 @@
...
+                    DefaultCacheKeyringSchema,
+                    DefaultCacheKeyringCollection,
+                    cacheFileName,
+                    DefaultCacheKeyringAttribute1,
+                    DefaultCacheKeyringAttribute2)
+                .Build();
+
+            var cacheHelper = await MsalCacheHelper.CreateAsync(storageProperties).ConfigureAwait(false);
+#pragma warning disable CS0618 // MsalCacheHelper.Clear is obsolete but is the correct approach for full cache wipe on disconnect
+            cacheHelper.Clear();
```
JonathanCrd, line 101: Why is `Clear()` the intentional approach here? This would wipe out the entire MSAL Cache, removing all accounts, not only the ones in the current context.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
```diff
@@ -430,6 +430,18 @@ public static async Task<IAuthContext> LogoutAsync()
...
+                }
+                catch (Exception)
+                {
```
JonathanCrd, line 465: Any actual failure here would be silent, could you add a log here so it's diagnosable?

JonathanCrd (commented):
Overall it looks good but I left a couple of comments about certain scenarios that might be missing and some questions too.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
gavinbarron: @JonathanCrd I think we probably need to clear the WAM token cache if possible to ensure a more complete fix.

src/Authentication/Authentication.Core/Utilities/TokenCacheUtilities.cs
gavinbarron: My thinking here is that it would be roughly equivalent to the approach of iterating over the results from `.GetAccounts()` to remove each or them, how does this differ given that the cache files are managed per machine user?

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
gavinbarron: Additionally, would using this approach avoid the need to have the TokenCacheUtilities class?

JonathanCrd: I don't think so, the broker only covers Windows, and there are non-broker auth flows on windows too.

src/Authentication/Authentication.Core/Utilities/TokenCacheUtilities.cs
JonathanCrd: I see. My comment was more about removing only the account being disconnected rather than all accounts, mainly from a UX perspective. That said, if there aren't any issues with clearing all accounts, then this approach works for me.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
gavinbarron: @JonathanCrd Sorry, I wasn't very clear here I was more meaning the general approach of using a PCA instance to clear the cached tokens not that specific implementation. Maybe something like this to provide broad cover and eliminate the additional code that knows about the caching specifics

```csharp
      var builder = PublicClientApplicationBuilder
              .Create(authContext.ClientId)
              .WithAuthority(GetAuthorityUrl(authContext));
      if (ShouldUseWam(authContext))
      {
          builder = builder
              .WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
      }
      var pca = builder.Build();
      foreach (var account in await pca.GetAccountsAsync().ConfigureAwait(false))
          await pca.RemoveAsync(account).ConfigureAwait(false);
```

src/Authentication/Authentication.Core/Utilities/TokenCacheUtilities.cs
gavinbarron: The CLI does not offer an account parameter, so any user account selection is done during the browser flow, which requires Disconnect-MgGraph to trigger the account selection again, so this feel like the right option. I suppose that there could be the case where a user has closed a terminal and is running Connect-MgGraph again triggering account selection.
But either way I'm comfortable with fully clearing out any cached tokens,

g2vinay: Msal cache scope is per window user, so .clear() will only scope to disconnecting user, I believe.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
g2vinay: TokenCacheUtilities -> ClearPersistedTokenCacheAsync still needs to be retained as it deletes the file based caches. PCA code above is needed on top to clear the WAM caches.

Just heads up
WAM is shared broker storage across Windows Azure tools. Removing accounts would also sign out:
Visual Studio
Azure CLI
Azure PowerShell
VS Code Azure extensions
Any app using Azure.Identity with broker enabled

This can impact UX for other logged in flows.

g2vinay:
Looks fine to me, left my comments on the ongoing discussions.

src/Authentication/Authentication.Core/Utilities/AuthenticationHelpers.cs
gavinbarron: @g2vinay does the above code impact all those other apps given that the PCABuilder is specifically targeting the ClientId for this Module?

g2vinay: Using a module specific ClientId isolates app specific token entries, but once .WithBroker(...) is enabled, account state comes from WAM (Windows broker), which is shared at the OS level. Calling pca.RemoveAsync(account) removes that broker account session, so other broker-enabled apps using the same signed-in Windows account may lose silent SSO and prompt for sign-in again.

You can double check this behavior with a simulated setup.

That cross app impact is why we should only use broker account removal when we explicitly want a full sign-out experience and are comfortable with the broader UX tradeoff.

gavinbarron: Personally, I don't think that we should be having such widespread impact to other applications that the customer is using. 

To this end I think we need to add an additional flag to the `Disconnect-MgGraph` cmdlet to allow a user to opt into this behavior. Perhaps `-RemoveWAMCachedTokens`

g2vinay: Opt-In model with documentation works well.

gavinbarron: Added `SignOutFromBroker` as a SwitchParameter

ramsessanchez (approved):

PR #3653, opened 2026-06-30
fix: bootstrap rush/autorest prerequisites in GenerateModules.ps1 for fresh clones

ramsessanchez:
## Problem
Running `tools/GenerateModules.ps1` on a fresh clone fails at:
```
Command 'npx --no-install rush install' failed with exit code 1.
```
`npx --no-install` resolves `rush`/`autorest` only from the repo-root `node_modules`, which does not exist on a fresh local clone. CI is unaffected because `.azure-pipelines/common-templates/install-tools.yml` installs these packages first, which masked the gap for local runs.

## Changes
- **Idempotent Node build-tools bootstrap** in `GenerateModules.ps1`: installs `@microsoft/rush`, `autorest@3.7.2` and `@autorest/core@3.10.4` at the repo root **only when the executables are missing** (same versions as `install-tools.yml`).
- **Fail-fast submodule check**: throws a clear `git submodule update --init --recursive` message if `autorest.powershell/rush.json` is absent, instead of a cryptic rush error.
- **.gitignore**: ignore the npm-generated root `package.json` / `package-lock.json`.

## Azure DevOps impact: none
The bootstrap is guarded by a presence check. In ADO the packages are already installed at the build sources directory, so the step is **skipped** - no new npm calls (safe for network-isolated release pipelines) and CI behavior is identical.

## Validation
- Fresh-clone path (no `node_modules`): installs tools, `rush`/`autorest` bins present, `rush install` then succeeds.
- Re-run / ADO path (tools present): bootstrap skipped.
- Missing-submodule path: throws the actionable error.
- Script parses with 0 errors; generated `package.json`/`node_modules` confirmed git-ignored.


Joywambui-maina:
Works locally on my end.

gavinbarron (approved):

PR #3658, opened 2026-07-13
fix: unwrap PSObject-wrapped values when serializing Invoke-MgGraphRequest dictionary bodies

### LINKED ISSUE #3654: Invoke-MgGraphRequest -Body serialization fails on PSObject-wrapped pipeline strings
### Describe the bug

`Invoke-MgGraphRequest -Body` serializes `IDictionary` bodies with Newtonsoft.Json. If a hashtable body contains a value that came from a PowerShell pipeline (for example, bare `$_` from `ForEach-Object`) Newtonsoft sees a `System.Management.Automation.PSObject` wrapper rather than a plain CLR string and reflects into PowerShell adapted members on the string.

For a string value, this exposes the `Chars` indexed property as a `PSParameterizedProperty`. Serialization then fails with a circular/self-reference error before the HTTP request can be sent.

Observed exception:

```text
Newtonsoft.Json.JsonSerializationException: Self referencing loop detected for property 'Value' with type 'System.Management.Automation.PSParameterizedProperty'. Path 'message.toRecipients[0].emailAddress.address.Chars'.
```

This is surprising because from PowerShell the value still reports as `System.String`:

```powershell
$payload.message.toRecipients[0].emailAddress.address.GetType().FullName
# System.String
```

But from .NET, the dictionary value is actually a `System.Management.Automation.PSObject` whose `BaseObject` is a `System.String`.

### Expected behavior

`Invoke-MgGraphRequest -Body` should serialize idiomatic PowerShell hashtable bodies containing pipeline-produced primitive values the same way it serializes directly assigned primitive values.

At minimum, string-like `PSObject` wrappers should be unwrapped before JSON serialization so callers do not need to know about this PowerShell-to-.NET interop boundary.

### How to reproduce

The public `Invoke-MgGraphRequest` path requires an authenticated Graph context before it reaches body serialization. The following repro invokes the same internal `SetRequestContent(HttpRequestMessage, IDictionary)` path via reflection so it can be run without a tenant or token.

```powershell
Import-Module Microsoft.Graph.Authentication

$type = [Microsoft.Graph.PowerShell.Authentication.Cmdlets.InvokeMgGraphRequest]
$method = $type.GetMethod(
    'SetRequestContent',
    [System.Reflection.BindingFlags]'NonPublic,Instance',
    $null,
    @([System.Net.Http.HttpRequestMessage], [System.Collections.IDictionary]),
    $null)
$cmdlet = [Activator]::CreateInstance($type)

$toAddresses = @('recipient@example.com' -split '[;,]' |
    ForEach-Object { $_.Trim() } |
    Where-Object { $_ })

# This is the problematic, idiomatic PowerShell shape:
$body = @{
    message = @{
        toRecipients = @(
            $toAddresses | ForEach-Object {
                @{ emailAddress = @{ address = $_ } }
            }
        )
    }
    saveToSentItems = $true
}

$request = [System.Net.Http.HttpRequestMessage]::new(
    [System.Net.Http.HttpMethod]::Post,
    'https://graph.microsoft.com/v1.0/users/sender@example.com/sendMail')

$method.Invoke($cmdlet, @($request, $body))
```

Actual result:

```text
Exception calling "Invoke" with "2" argument(s): "Self referencing loop detected for property 'Value' with type 'System.Management.Automation.PSParameterizedProperty'. Path 'message.toRecipients[0].emailAddress.address.Chars'."
```

If the address assignment is changed to explicitly cast the pipeline value, serialization succeeds:

```powershell
@{ emailAddress = @{ address = [string]$_ } }
```

With that cast, the serialized body is:

```json
{"saveToSentItems":true,"message":{"toRecipients":[{"emailAddress":{"address":"recipient@example.com"}}]}}
```

A small probe showing the type mismatch at the .NET boundary:

```powershell
Add-Type -ReferencedAssemblies ([System.Management.Automation.PSObject].Assembly.Location) -TypeDefinition @'
using System;
using System.Collections;
using System.Management.Automation;
public static class DictProbe {
    public static string DescribeAddress(IDictionary body) {
        var message = (IDictionary)body["message"];
        var recipients = (IEnumerable)message["toRecipients"];
        object firstRecipient = null;
        foreach (var item in recipients) { firstRecipient = item; break; }
        var recipient = (IDictionary)firstRecipient;
        var email = (IDictionary)recipient["emailAddress"];
        var address = email["address"];
        var s = "dotnetType=" + address.GetType().FullName;
        if (address is PSObject p) {
            s += "; baseType=" + p.BaseObject.GetType().FullName;
            s += "; charsMemberType=" + p.Members["Chars"].GetType().FullName;
            s += "; charsMemberKind=" + p.Members["Chars"].MemberType;
        }
        return s;
    }
}
'@

[DictProbe]::DescribeAddress($body)
```

Output for the failing body:

```text
dotnetType=System.Management.Automation.PSObject; baseType=System.String; charsMemberType=System.Management.Automation.PSParameterizedProperty; charsMemberKind=ParameterizedProperty
```

Output after using `[string]$_`:

```text
dotnetType=System.String
```

### SDK Version

Reproduced with:

```text
Microsoft.Graph.Authentication 2.25.0
Microsoft.Graph.Authentication 2.35.1
```

The current source also appears to still call Newtonsoft directly for dictionary bodies:

```csharp
// InvokeMgGraphRequest.SetRequestContent(HttpRequestMessage request, IDictionary content)
var body = JsonConvert.SerializeObject(content);
```

None known.

### Known Workarounds

Explicitly cast pipeline-derived primitive values before placing them into the hashtable body, e.g.:

```powershell
$toAddresses | ForEach-Object {
    @{ emailAddress = @{ address = [string]$_ } }
}
```

or otherwise ensure values crossing into `Invoke-MgGraphRequest -Body` are plain CLR primitives rather than `PSObject`-wrapped values.

### Debug output

No Graph debug output is produced in the repro above because the failure happens during request content serialization, before the HTTP request is sent.

### Configuration

Local repro environment:

```text
PowerShell 7.6.3
Debian GNU/Linux 13 (trixie)
x64
```

### Other information

This looks like a PowerShell/.NET interop robustness issue in `Invoke-MgGraphRequest`'s body serializer rather than a Graph API issue. PowerShell callers can construct a normal-looking hashtable where `.GetType()` reports `System.String`, but a generic .NET serializer sees a `PSObject` wrapper and walks PowerShell adapted members such as `Chars`.

Related prior art for this general failure class exists outside the Graph SDK, e.g. serialization of raw PowerShell objects hitting circular reference errors involving `PSParameterizedProperty`.

Virgil-Bulens:
Fixes #3654

- Add `PSObjectJsonConverter`, a Newtonsoft `JsonConverter` that unwraps `PSObject`-wrapped values to their `BaseObject` during serialization, at any nesting depth. Pure `PSCustomObject`s (which have no underlying CLR object) are projected into JSON objects from their properties.
- Pass the converter at the single `JsonConvert.SerializeObject` call site in `InvokeMgGraphRequest.SetRequestContent(HttpRequestMessage, IDictionary)`, so idiomatic hashtable bodies containing pipeline-produced values (bare `$_` in `ForEach-Object`) serialize the underlying CLR values instead of failing with `Self referencing loop detected for property 'Value' with type 'System.Management.Automation.PSParameterizedProperty'`.
- Add xUnit tests covering PSObject-wrapped strings/primitives/dictionaries, `PSCustomObject` bodies, and the `SetRequestContent` wiring (made `internal`; the assembly already has `InternalsVisibleTo` for the test project).
- Pin the test project's `Microsoft.PowerShell.SDK` reference to 7.4.17. 7.5.x targets net9.0 only and contributes no assemblies to the net8.0 test build, which silently left the `PowerShellStandard.Library` stub (whose APIs return null) as the runtime `System.Management.Automation`. This is why the existing `ShouldReturnPsObject` tests fail on net8.0, and it made the new PSObject-dependent tests impossible to run. With the pin, the pre-existing `StringUtilTests.ShouldReturnPsObject` failures on net8.0 are also resolved.

### Verification

- `dotnet test src/Authentication/Authentication.Test -f net8.0`: 84/85 pass on Linux; the one failure (`AuthenticationHelpersTests.ShouldThrowIfNonExistentCertNameIsProvidedAsync`) is pre-existing on an unmodified checkout of `main` and unrelated (certificate store).
- The exact repro from #3654 was run in PowerShell 7.6 against the patched assembly and now produces:
  ```json
  {"message":{"toRecipients":[{"emailAddress":{"address":"recipient@example.com"}}]},"saveToSentItems":true}
  ```
- A nested `[pscustomobject]@{ address = "..." }` body value, which previously failed the same way, serializes to `{"emailAddress":{"address":"recipient@example.com"}}`.

### Other links

- https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/3654


Virgil-Bulens:
@microsoft-github-policy-service agree

ramsessanchez (approved):

ramsessanchez (approved):

PR #3675, opened 2026-07-23
Add custom xtap cmdlets

rajaramkd:
﻿

Fixes #

-
-
-

### Other links

-
-
-


src/Identity.SignIns/beta/custom/New-PartnerEntraXTAP.ps1
```diff
@@ -0,0 +1,131 @@
...
+        ${M365CollaborationOutboundTargetGroups},
+
+        [Parameter(ParameterSetName = 'Expanded')]
+        [ValidateSet('Allowed', 'Blocked')]
+        [System.String]
+        ${AppServiceConnectInbound},
+
+        [Parameter(ParameterSetName = 'Expanded')]
+        [System.String[]]
+        ${AppServiceConnectInboundTargetApplications},
```
sathyas8, line 47: All ParameterSetName are set to "Expanded", don't you want to use different parameter sets for m365Collab and AppServicesConnect?

rajaramkd: We merged the two parameter sets so that M365Collab and AppServiceConnect can be configured in a single call, consistent with the Graph API, which allows both trust flags to be updated together.

PR #3682, opened 2026-07-29
Add WrapperGenerator: a standalone cmdlet generator that reproduces published MS Graph cmdlet names

Joywambui-maina:
### Changes Proposed

- Adds `tools/WrapperGenerator`, a standalone .NET CLI that generates C# cmdlet
  classes (e.g. `Get-MgUserMessage`) from Graph's OpenAPI description
- Builds cmdlet nouns from URL paths with per-word singularization instead of
  operationIds, for deterministic naming
- Reproduces published SDK hand-tuned names as checked-in data, with entries citing their AutoRest directive sources
- Emits paired list/item GETs as one public dispatcher cmdlet with `List`/`Get`
  parameter sets forwarding to internal cmdlets
- Includes 69 unit tests validated against published Microsoft.Graph cmdlet
  names from `MgCommandMetadata.json`
- Adds `tools/Compare-WrapperCmdletNames.ps1` for parity validation against the command metadata inventory
- Documents the naming algorithm rules with shipping cmdlets as evidence, plus identified gaps
- Updates `.azure-pipelines/common-templates/install-tools.yml` to install the
  .NET 10 SDK so CI can build the `net10.0` generator projects

### Why

Customer scripts depend on the exact cmdlet names the SDK ships — `Get-MgUserMessage`, not `Get-MgUsersMessages`. When nouns are derived from operationIds, generated names leak whatever plurality the spec author chose, and collisions have occurred. Path-based deterministic naming with singularization rules keeps name parity stable.

### Validation

- All 69 unit tests pass (`dotnet test` on net10.0).
- Generated modules were checked against the published cmdlet inventory
  (`MgCommandMetadata.json`) using `tools/Compare-WrapperCmdletNames.ps1`: every generated cmdlet whose API call exists in the published SDK matches the published name exactly (e.g. BackupRestore 183/186, Bookings 129/131). The handful of non-matches are endpoints with no published cmdlet to compare against — not naming mismatches.
- The generated C# compiles alongside the Kiota-generated client; module
  wiring is documented as future work.


Joywambui-maina:
@microsoft-github-policy-service agree company="Microsoft"

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+    IReadOnlyList<HeaderParam> HeaderParams);
+
+public static class Naming
+{
+    // GET->Get, POST->New, PATCH->Update, PUT->Set, DELETE->Remove (design spec section 7).
+    // The attribute class is tracked too because it differs per verb: Update lives in
+    // VerbsData, the rest in VerbsCommon.
+    private static readonly Dictionary<string, (string VerbsClass, string VerbName)> VerbMap = new(StringComparer.OrdinalIgnoreCase)
+    {
+        ["get"] = ("VerbsCommon", "Get"),
```
peombwa: Use strongly-typed values here instead of raw strings for both the HTTP method key and the cmdlet verb name - e.g. `System.Net.Http.HttpMethod` for the key and a `PSVerb`/enum-like wrapper (or at least a constant) for the verb, rather than comparing/storing "get", "Get", etc. as bare strings.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(operation);
+        if (!VerbMap.TryGetValue(operation.HttpMethod, out var verb))
+            throw new NotSupportedException($"No cmdlet verb mapping for HTTP method '{operation.HttpMethod}'.");
+
+        // The noun comes from the URL path, not the operationId. OperationIds keep whatever
+        // plurality the spec author chose, while the published SDK names follow the path:
+        // GET /users/{id}/messages is Get-MgUserMessage. The few hand-tuned exceptions the
+        // published SDK carries are mirrored as data in NamingOverrides, never as code here.
+        var noun = "Mg" + NamingOverrides.ApplyNounOverrides(operation.HttpMethod, operation.Path, BuildNounFromPath(operation.Path));
```
peombwa: Move the hard-coded `"Mg"` prefix into a constants class rather than inlining it here - we'll want this to be a user-configurable value later, and centralizing it now avoids a scattered find/replace.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        // A list GET (/users/{id}/messages) and its item GET (/users/{id}/messages/{message-id})
+        // get the same noun on purpose. PowerShellWrapperGenerationService pairs them into one
+        // public Get-MgX dispatcher cmdlet; the two real implementations get suffixed names via
+        // WithSuffix below.
+        var className = $"{verb.VerbName}{noun}Command";
```
peombwa: Why does the generated class name get a `Command` suffix? If it is needed, please add a comment explaining the convention (e.g. is this to avoid a name clash with the cmdlet's own noun/verb, or just a project convention?), since it isn't obvious from the surrounding code.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+        };
+    }
+
+    // Pascal-cases and singularizes every fixed path segment, then joins them. Two kinds of
+    // repetition are dropped, because the published SDK drops them:
+    //   /sites/{id}/sites                  -> Site (not SiteSite)
+    //   /domains/{id}/domainNameReferences -> DomainNameReference (the shared word "Domain"
+    //                                         appears once, matching Get-MgDomainNameReference)
+    // An OData cast segment like graph.user becomes AsUser, matching Get-MgGroupOwnerAsUser.
+    // Cast type names are already singular, so they skip the singularizer.
```
peombwa: The comment states cast type names are "already singular" - that's not guaranteed for every possible cast segment. Let's harden the code to singularize cast names in all cases rather than relying on that assumption.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+            var castType = segment.StartsWith("microsoft.graph.", StringComparison.OrdinalIgnoreCase)
+                ? segment["microsoft.graph.".Length..]
+                : segment.StartsWith("graph.", StringComparison.OrdinalIgnoreCase)
+                    ? segment["graph.".Length..]
+                    : null;
+            if (castType is not null)
+            {
+                parts.Add("As" + castType.ToFirstCharacterUpperCase());
+                continue;
+            }
```
peombwa: Let's extract this block into a dedicated helper method (e.g. TryBuildCastSegmentNoun) so the cast-segment logic is separated from the general segment-walking loop and can be unit tested in isolation.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            if (segment.StartsWith('{') && segment.EndsWith('}'))
+                names.Add(segment[1..^1].ToPascalCase('-'));
+        }
+        return names;
+    }
+
+    // The Kiota-generated ApiClient exposes a property per fixed path segment and an indexer
+    // per path parameter, e.g. client.Users[UserId].Messages[MessageId].
+    private static string BuildBuilderExpression(string path, List<string> pathParamNames)
```
peombwa, line 161: Please add a comment describing what this method does and the shape of the string it returns so callers don't have to reverse-engineer it from the body.

tools/WrapperGenerator/CmdletNaming.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    // A fixed path segment becomes one Kiota request-builder member. An OData cast segment
+    // carries dots ("microsoft.graph.user", or "graph.user" in the KiotaCompat specs); Kiota
+    // maps it to a single member by upper-casing each dot-separated part and concatenating, so
+    // "microsoft.graph.user" is the "MicrosoftGraphUser" property, "graph.user" is "GraphUser"
+    // (verified against a generated Kiota C# client). Emitting the raw segment would leave a
+    // stray dot ("Graph.user") — invalid C# and not a real builder member. Non-cast segments
+    // have no dots and are unchanged. NOTE: cast endpoints are not generated end to end yet
+    // (tracked follow-up), so this keeps the expression a valid identifier chain until then.
+    private static string ToBuilderMemberName(string segment) =>
```
peombwa: This method's logic only really applies to OData cast segments - the name should reflect that (e.g. `ToCastAwareBuilderMemberName` or similar) so it's self-documenting instead of reading as a generic segment-to-member converter.

tools/WrapperGenerator/PowerShellWrapperGenerationService.cs
```diff
@@ -0,0 +1,370 @@
...
+    {
+        entityTypeName = string.Empty;
+        var id = schema.GetReferenceId();
+        if (string.IsNullOrEmpty(id))
+            return false;
+        entityTypeName = SchemaNameToTypeName(id, modelsNamespace);
+        return true;
+    }
+
+    private static string ResolveReferenceId(IOpenApiSchema schema) =>
```
peombwa: `ResolveReferenceId` is dead code and is never called anywhere AFICT. Please remove it.

.gitignore
```diff
@@ -61,6 +61,7 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+sweep_results.json
```
peombwa: Orphaned `.gitignore` entry. No file produces in this diff.

.azure-pipelines/common-templates/install-tools.yml
```diff
@@ -5,7 +5,12 @@ steps:
   - task: UseDotNet@2
     displayName: Use .NET SDK
```
peombwa: nit: 
```suggestion
    displayName: Use .NET 10 SDK
```

.azure-pipelines/common-templates/install-tools.yml
```diff
@@ -5,7 +5,12 @@ steps:
...
     inputs:
-      version: 8.x 
+      version: 10.x
+
+  - task: UseDotNet@2
+    displayName: Use .NET SDK
```
peombwa: nit:

```suggestion
    displayName: Use .NET 8 SDK
```

peombwa (commented):
Nice work on this - the naming/singularization engine looks good to me for a first pass, and pinning tests to real published cmdlet names instead of arbitrary expectations was the right call. I'm yet to live run the branch, but I'll more comments later.

Please see my comments below for areas to address in the meantime.

tools/WrapperGenerator/Singularizer.cs
```diff
@@ -0,0 +1,115 @@
...
+//
+// This is not a general English inflector. The vocabulary is the finite set of Graph path
+// segments, and every rule exists because a shipping cmdlet needs it. The README in this
+// folder lists each rule with the cmdlet that proves it.
+//
+// One behavior matters most: the published SDK singularizes every camel-case word in a
+// segment, not just the last one. "termsAndConditions" ships as TermAndCondition and
+// "onPremisesSynchronization" as OnPremiseSynchronization. SingularizeSegment does the same:
+// it splits a segment into words and runs the rules on each word.
+public static partial class Singularizer
```
peombwa, line 18: Let's create a follow up issue to look into using external libraries such as [Humanizer](https://github.com/Humanizr/Humanizer) to handle the singularization to ensure this scales.

tools/WrapperGenerator/SchemaProperties.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        Walk(schema);
+        return result;
+    }
+
+    // passwordProfile is a nested complex type, so ExtractPrimitiveProperties skips it, but
+    // Graph requires it to create a user. This flag lets the emitter add the two flattened
+    // parameters (-Password, -ForceChangePasswordNextSignIn) that make New-MgUser usable.
+    public static bool HasPasswordProfile(IOpenApiSchema schema)
```
peombwa, line 51: Create a follow-up issue to explore how the `passwordProfile` pattern (special-casing complex types to flatten their top-level properties as parameters) could be applied to other complex types in the schema. This would expand cmdlet usability while maintaining the current shallow-binding approach.

For now, `passwordProfile` is the only known required complex type, so it's reasonable to special-case it here as a temp. A tracked follow-up issue will ensure this pattern doesn't become accidental technical debt if other complex types need similar treatment later.

tools/WrapperGenerator.Tests/WrapperGenerator.Tests.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
peombwa: This is a good suggestion. Please remove the OutputType line from the test csproj.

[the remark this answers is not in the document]
Joywambui-maina: Done in 8b49c5e — all 78 tests still pass. One note for the record: nothing was failing to compile (xunit.v3 auto-generates the entry point for exe builds), but the line was indeed unnecessary since dotnet test runs through the VSTest adapter either way.

tools/WrapperGenerator/Singularizer.cs
Joywambui-maina: Tracked in #3689

tools/WrapperGenerator/SchemaProperties.cs
Joywambui-maina: Tracked in #3690

peombwa (approved):

PR #3694, opened 2026-08-04
fix(wrapper-generator): singularizer corrections and naming edge-case catalog

Joywambui-maina:

- Correct five singularization words found by auditing the naming rules against the full  `MgCommandMetadata.json` inventory: `cookies`, `skus`, `dns`, `ios`, and `statistics` (the last surfaced by the DEVX API's Humanizer exception list) — each pinned by a test citing the shipped cmdlet.
- Deliberately correct `Get-MgSecurityThreatIntelligenceHostWhoi` → `...HostWhois`: only 2 of the 30 shipped whois-family commands truncate "Whois", an AutoRest inflection defect. Per review discussion, corrected names ship with no alias for the old name. The parity gate gains a small allowlist and reports these as `[CORRECTED]` instead of failing.
- Start the naming edge-case catalog (`tools/WrapperGenerator/edge-cases/naming-edge-cases.md`): one Markdown file per class of issue with fixed entry fields.
- Map numeric parameter types by OpenAPI `format` so `number`+`int32/int64` properties emit `int`/`long` matching the Kiota client's model types instead of `double`.


tools/WrapperGenerator/docs/edge-cases/naming-edge-cases.md
peombwa, line 1: Please move the edge-cases folder under a docs folder - `tools\WrapperGenerator\docs\edge-cases`.

tools/WrapperGenerator/Singularizer.cs
```diff
@@ -18,17 +18,28 @@ namespace WrapperGenerator;
 public static partial class Singularizer
 {
     // Irregular plurals the SDK singularizes: Get-MgDriveItemChild, Get-MgUserPerson.
+    // "Cookies" would hit the ies-rule ("Cooky") but ships as Get-MgSecurityThreatIntelligenceHostCookie;
+    // "Skus" would hit the us-guard (stay put) but ships as Get-MgSubscribedSku.
```
peombwa: Let's keep comments focused on what properties/methods generally do rather than explaining individual cases. The same applies to the lines below and `SchemaProperties.cs` class.

tools/Compare-WrapperCmdletNames.ps1
```diff
@@ -126,6 +131,17 @@ function Get-ModuleApiVersion {
     return $null
 }
 
+# Published names the generator deliberately corrects instead of reproducing. Each entry maps
+# the shipped (wrong) command to the corrected one the generator emits, and must have a matching
+# entry in tools/WrapperGenerator/edge-cases/naming-edge-cases.md and a pinned naming test. The
+# gate reports these as [CORRECTED] instead of [MISMATCH] and does not fail on them.
+$deliberateCorrections = @{
```
peombwa, line 138: Let's avoid manually hardcoding drifts as it won't scale give the size of MS Graph. We can look into automating this by:
- Generate/consume the current SDK's output (old code + old rules)
- Generate the new SDK's output (new code + new rules)
- Diff the two to find all cmdlets that changed names

This can be a follow up item.

tools/WrapperGenerator/Singularizer.cs
peombwa, line 76: The singularization logic currently relies on a long guard chain that makes it difficult to test individual rules and extend them without risking regressions. This change adds more hardcoded corrections, which increases the maintenance burden.

Consider refactoring this into a rule-based pattern using strategy objects - each singularization rule as its own strategy implementation. This would:
- Isolate each rule for independent testing
- Make it clearer what each rule applies to
- Simplify adding new corrections without modifying a monolithic guard chain

I've created issue https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/3699 to track this.

peombwa (commented):
The changes LGTM. I've left a couple of comments for you to look at. The rule-based refactor and automated comparison can be done as follow up items. Please ensure that we have corresponding tracking issues for them.

tools/WrapperGenerator/docs/edge-cases/naming-edge-cases.md
Joywambui-maina: Both comments addressed. Follow-ups tracked: #3699 and #3700

peombwa (approved):

Issue #912, opened 2021-10-21
Investigate how to mitigate AutoREST command name truncation when operationIds contain prepositions


peombwa:
Given this https://github.com/Azure/autorest.powershell/issues/795 limitations. We will need to figure out how to modify our operationIds to avoid prepositions from being truncated in command names by AutoREST. e.g.,

| OperationId | Generated Command | Expected Command |
| - | - | - |
| reports_getSkypeForBusinessActivityCounts | Get-MgReportSkype | Get-MgReportSkypeForBusinessActivityCount  |
| groups_ListAcceptedSendersByRef | Get-MgGroupAcceptedSender | Get-MgGroupAcceptedSenderByRef |
| applications_GetCreatedOnBehalfOfByRef | Get-MgApplicationCreatedOnBehalf | Get-MgApplicationCreatedOnBehalfOfByRef |
| applications_GetCreatedOnBehalfOf | Get-MgApplicationCreatedOnBehalf (conflicts with above) | Get-MgApplicationCreatedOnBehalfOfByRef |

PR #3698, opened 2026-08-06
feat(wrapper-generator): module packaging scripts and full v1.0 validation

Joywambui-maina:

- **Packaging scripts**: `Build-WrapperModule.ps1` turns an OpenAPI doc into an importable module (kiota client + wrappers + dll + PSD1); `Test-WrapperModule.ps1` imports it in a fresh pwsh and smoke-tests it like a user would. Generation defaults to the
  Kiota-compatible docs — kiota rejects or hangs on the PowerShell-profile ones.
- **Eight emitter fixes**, each found by compiling a real module and pinned by a test: worker errors no longer masked by dispatchers, `-DeviceId1`-style collision handling matching the published SDK, fully-qualified model types mirroring kiota's renames, query options only where declared, re-fetch only where a GET exists, media endpoints skipped.
- **Two naming findings**: shipped `*-MgPlaceCheck` is an AutoRest preposition truncation (#912 class) — corrected per policy; "Rights" added to the invariants (`Get-MgPrivacySubjectRightsRequest`), taking Compliance to 23/23.
- **Catalog**: new `edge-cases/kiota-alignment-edge-cases.md` documenting each defect class.

**Results:** 35/35 cmdlet-producing v1.0 modules build and pass import smoke tests (the 4 actions/functions-only modules emit nothing by design); names match the published inventory wherever it has an entry, corrections reported as `[CORRECTED]`. 103 tests pass.


tools/Build-WrapperModule.ps1
```diff
@@ -0,0 +1,228 @@
...
+$cmdletAttrPattern = '\[Cmdlet\(Verbs\w+\.(\w+),\s*"((?:\\.|[^"\\])*)"'
+function Get-EmittedCmdletNames {
+    param([string]$CmdletsDir)
+    Get-ChildItem -Path $CmdletsDir -Filter '*.g.cs' -File | ForEach-Object {
+        $match = [regex]::Match((Get-Content -Path $_.FullName -Raw), $cmdletAttrPattern)
+        if ($match.Success) {
+            "$($match.Groups[1].Value)-$([regex]::Unescape($match.Groups[2].Value))"
+        }
+    }
+}
```
peombwa: This logic in lines 85 - 96 parses the file content to extract cmdlet names. This is slow, fragile, and detached from the compilation logic.

### Solution
Instead of inferring cmdlets by parsing generated files, align cmdlet discovery with the csproj structure:

1. Have WrapperGenerator output cmdlets to a dedicated `Cmdlets` folder (already does this at line 121)
2. The csproj explicitly includes `Cmdlets/**/*.g.cs`:
``` xml
<ItemGroup>
    <Compile Include="{CmdletSourcePath}\**\*.g.cs" />
  </ItemGroup>
```
3. Extract cmdlets count directly from the compiled assembly or read a generated manifest thus eliminating the need for `Get-EmittedCmdletNames`.

tools/Build-WrapperModule.ps1
```diff
@@ -0,0 +1,228 @@
...
+
+function Build-OneModule {
```
peombwa: nit: Have the function name follow verb-noun convention. The one can be inferred from the parameter name.

```suggestion
function Build-Module {
```

tools/Build-WrapperModule.ps1
```diff
@@ -0,0 +1,228 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Kiota.Bundle" Version="2.0.0" />
+    <PackageReference Include="PowerShellStandard.Library" Version="5.1.1" PrivateAssets="all" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$authCsproj" />
+  </ItemGroup>
+
+</Project>
```
peombwa: Instead of hardcoding the csproj content as here-strings in the script, move the templates to dedicated files in `tools/Templates/` and use find-and-replace patterns. This follows the existing repo pattern (see `FileUtils.ps1` and `GenerateServiceModule.ps1`).

tools/Build-WrapperModule.ps1
```diff
@@ -0,0 +1,228 @@
...
+    <NoWarn>`$(NoWarn);CS1591</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Kiota.Bundle" Version="2.0.0" />
+    <PackageReference Include="PowerShellStandard.Library" Version="5.1.1" PrivateAssets="all" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$authCsproj" />
```
peombwa: The csproj doesn't reference the generated Kiota client. How will the Client code be compiled into the output?

The recommended approach should be:
1. Generate a Client.csproj for the Kiota output
2. Reference it from the wrapper's .csproj as a <ProjectReference>
3. Compile both together

peombwa (dismissed):

tools/Build-WrapperModule.ps1
Joywambui-maina: Addressed in dec945a. The wrapper project now explicitly compiles Cmdlets/**/*.g.cs, and cmdlet names are discovered from the compiled module by importing its DLL in a fresh PowerShell process.

tools/Build-WrapperModule.ps1
Joywambui-maina: Addressed. The Kiota output now gets a dedicated Client/Client.csproj. The wrapper project explicitly compiles the generated cmdlets and references both the Client project and Authentication project, so they build together through project references.

ramsessanchez:
What is the plan for the NamingOverides file? I see these as similar to directives, but does this require us to name each file separately?

ramsessanchez (approved):

Issue #3704, opened 2026-08-10
WrapperGenerator: fail loudly when two cmdlets resolve to the same file name


Joywambui-maina:
Summary
The generator silently overwrites cmdlet files when two operations resolve to the same cmdlet name, which drops cmdlets without any error, the same failure mode AutoRest had.

What's working now
One `.g.cs` file is written per cmdlet and for almost every path this is correct. Modules build and import cleanly, and the parity gate verifies the names of the files that survive.

How can we make it better
When two operations produce the same cmdlet name, the second file overwrites the first with no warning. Verified in the Sites module: `GetMgSite.g.cs` contains the sub-sites call (`client.Sites[SiteId].Sites[SiteId1]`), because it overwrote the plain get-site-by-id cmdlet. No test or gate can catch this today since the surviving file compiles and imports.

The fix is two steps: track written file names and fail generation on any collision, then resolve the known collisions with rename entries (the published SDK ships `Get-MgSubSite` for this exact case).

Joywambui-maina:
The collision guard and the first batch of resolutions are up in a PR (stacked on #3698).

Joywambui-maina:
- Make cmdlet name extensible (directives)

Joywambui-maina:
Update: the collision work is now complete on PR #3713.

The original fail-loud guard remains the backstop: generated file names are tracked case-insensitively, and generation fails with all detected collisions instead of allowing a later operation to overwrite an earlier cmdlet silently.

I ran the guard across all 39 v1.0 modules and used the shipped command metadata as the resolution oracle. Resolutions are derived by exact API version, HTTP method, and normalized URI rather than broad naming heuristics:
- routes for which the published SDK ships no command are explicitly suppressed;
- routes with distinct published commands receive the corresponding published rename;
- the Sites collision now preserves both operations, including `Get-MgSubSite`;
- the checked-in resolution data is reproducibly derived from the collision inventory and command metadata;
- a drift test fails if the checked-in suppression or rename data no longer matches a fresh derivation.

Validation results:
- 39/39 v1.0 modules generated;
- zero unresolved cmdlet file collisions;
- 20 commands previously lost to silent overwrites were recovered;
- exact-name parity improved by 56 commands;
- 121/121 WrapperGenerator tests pass;
- collision regression tests prove that an unhandled duplicate still fails loudly and reports both conflicting operations.

[...]

PR #3713, opened 2026-08-11
fix(wrapper-generator): fail loudly on cmdlet file collisions

Joywambui-maina:

- Fail generation loudly when two operations resolve to the same cmdlet file, listing every collision with both operations' request-builder expressions, instead of letting the later one silently overwrite the earlier. This is the silent-drop failure mode AutoRest had, and nothing could detect it before: writes are not logged at console level, the summary counts surviving files and the parity gate only inspects files that exist.
- Merge OData cast list/item pairs (owners/graph.user with owners/{id}/graph.user) into one Get-* dispatcher the way plain list/item pairs already merge, matching the single cmdlet the published SDK ships for them.
- Resolve the collisions the full-inventory sweep surfaced as NamingOverrides data: the termStore and agreement-file stitches, the default-singleton renames (SubSite, DefaultDrive, DefaultCalendarEvent, GroupLifecyclePolicyByGroup) and the nested navigations the SDK never shipped. Every entry cites the AutoRest directive or the published command inventory that proves it.
- Add PathMatch.Suffix to NamingOverrides so a navigation that recurs under many roots takes one entry instead of one per root.
- Surface the generator's exception instead of its skip warnings when Build-WrapperModule.ps1 captures a failure, so a collision failure is readable in the build summary.
- Document both collision classes in the edge-case catalog, and pin the renamed names, the suppression matching and the guard itself in the unit tests (115 passing).
- Six modules now fail generation instead of building green: Calendar 3, Files 1, Groups 8, Identity.Governance 80, Notes 18 and Sites 102 collisions. That is the guard working rather than a regression, because those modules were already losing cmdlets silently.

### Other links

- Part of #3704, which carries the remaining collision families with their evidence. Two of them need a design call before they can be encoded.


peombwa (commented):
@Joywambui-maina - great work here on detecting and including the collision inventory. I wanted to share some feedback on PR structure that'll help both you and the reviewers going forward.

This PR bundles several independently valuable changes into one 6,600-line diff, which makes it quite hard to review confidently. As a general practice, each PR should do **one thing** - a single, self-contained change that a reviewer can understand, evaluate, and approve (or request changes on) without needing to hold the entire context of unrelated changes in their head at the same time.

Looking at the description, I'd suggest this could have been split into roughly **four focused PRs**, merged in order to avoid conflicts:

1. **Collision guard** - the `writtenCmdletFiles` tracking in `PowerShellWrapperGenerationService.cs` and its regression test. This is the core safety net and could land first with zero naming-data dependencies.

2. **OData cast list/item pairing** - the `IsListItemPair` enhancement in `CmdletNaming.cs`, the `TrailingCastMember` helper, and its unit tests. A standalone behavior change with its own tests.

3. **Curated NamingOverrides expansion** - the `PathMatch.Suffix` model change, the ~30 new curated entries (SubSite, DefaultDrive, termStore stitching, Security nested navs, etc.), their naming tests, and the edge-case docs. This is the domain-knowledge-heavy part reviewers need to scrutinize most carefully.

4. **Derived collision-resolution pipeline** - `Derive-CollisionResolutions.ps1`, `DerivedCollisionResolutions.cs`, the embedded JSON data files, the drift-gate test, and the `--api-version` / `--no-collision-data` CLI flags. This is a whole new subsystem with its own derivation, validation, and embedding strategy.

Each of these tells a clear story, is independently testable, and is small enough that a reviewer can give it proper attention. When they're combined, it's easy for reviewers to rubber-stamp the data files (which are the bulk of the diff) instead of carefully evaluating whether each entry is correct - which defeats the purpose of code review.

No need to split this one retroactively - let's get it across the line as-is. But for future work, let's keep the PRs focused on just one thing.

Let's also address the copilot comments.

Issue #3705, opened 2026-08-10
WrapperGenerator: wire generated modules into src/{Module} with per-module namespaces


Joywambui-maina:
Summary
Generated modules are throwaway test builds under artifacts/; wire them into the real repo layout with per-module namespaces and a beta prefix so they can become shippable.

What's working now
Build-WrapperModule.ps1 builds any module into an importable PSD1 (kiota client + wrappers + dll + manifest), and all 35 cmdlet-producing v1.0 modules build and pass import smoke tests.

How can we make it better
Every module shares one hard-coded `MgPoC` cmdlet namespace, the generated csproj contains a machine-local absolute path and is marked do-not-commit, and nothing lands in src/{Module}/ where real modules live. There is also no `MgBeta` noun prefix, so beta modules cannot generate correct cmdlet names at all. Add a namespace/prefix option to the generator, emit a committable project layout under src/{Module}/, and decide the target framework for shipping.

Joywambui-maina:
### Already implemented

- Build-WrapperModule.ps1 generates both required source layers from the same OpenAPI document:
  - the Kiota client under `Client/`
  - the PowerShell wrapper cmdlets under `Cmdlets/`
- `-IntoSource` defines a source layout under `src/<Module>/v1.0/wrapper/`.
- Generated clients use module-specific namespaces.
- Generated wrapper projects use relative references to the client and Authentication projects.
- Client and wrapper project templates exist as standalone files.
- Wrapper-specific build, test, parity, and collision gates exist for artifact builds.
- The current KiotaCompat v1.0 input contains 38 module specifications, and local artifact output exists for all 38 modules.

### Partially implemented

- The script can emit a self-contained source directory, but this capability has not been exercised into the repository source tree.
- Standalone wrapper artifacts compile locally, but they are separate test modules named `Microsoft.Graph.Wrapper.<Module>`.
- The wrapper-specific gates validate artifact builds, but they are not part of the production build or packaging path.

### Not implemented

- There are currently zero tracked files under `src/*/v1.0/wrapper/`.
- Neither generated Kiota client sources nor generated cmdlet sources are committed under src.
- A clean checkout cannot build the wrappers without first running generation.
- Generation requires an arbitrary globally installed `kiota` executable; no repository-pinned Kiota tool manifest exists.
[...]

PR #3733, opened 2026-08-20
feat(wrapper-generator): commit v1.0 wrapper module sources under src…

Joywambui-maina:
Addresses #3705 — the committable layout under `src/{Module}/` and the target-framework decision. 

- Commit the generated Kiota client and wrapper cmdlets for all 35 cmdlet-producing v1.0 modules under `src/{Module}/wrapper/{ApiVersion}/`, so a clean checkout builds them with only the .NET SDK — kiota is needed to regenerate, never to compile.
- Nest the API version **under** `wrapper/`, not above it. AutoRest owns src/{Module}/{ApiVersion}/` and runs with `clear-output-folder`, so wrapper source placed there is deleted whenever the AutoRest modules regenerate — observed: generating Mail removed its wrapper tree. `src/{Module}/` itself is never cleared.
- Target `netstandard2.0` for both projects instead of `net10.0`. The compatibility contract comes from the module every wrapper references — src/Authentication/Authentication/Microsoft.Graph.Authentication.psd1` declares `PowerShellVersion 5.1` and `CompatiblePSEditions Core, Desktop` — and a manifest binds one un-conditioned dll path, so a single framework must satisfy both editions.
- Re-include the wrapper tree in `.gitignore`. Kiota derives directory names from Graph URL paths, so segments like `Publish/`, `Log/` and `BackupRestore/` collided with the existing Visual Studio rules and silently dropped 174 real source files.
- Record `kiota-lock.json`'s spec path relative to its output folder. It was absolute and machine-local, so every clone differed.
- Stop `DirectoryObjects` claiming `directory.publicKeyInfrastructureRoot`. The mapping gave the family to both `DirectoryObjects` and `Identity.DirectoryManagement`, emitting 11 identical public cmdlets from two modules; the published-command inventory assigns it to  `Identity.DirectoryManagement` alone.
- Repoint `Test-BodyBindingCoverage.ps1` and `Compare-WrapperOperationInventory.ps1` at the committed corpus, and replace the inventory's newest-write-time generator stamp with a SHA-256 over the generator's sources, embedded data and project file.
- Add `docs/WrapperCmdlets-V1.0.csv` and `-Summary.csv` as the reviewable inventory.

### How to review this

34,443 files, 34,427 of them generated. **Read `docs/WrapperCmdlets-V1.0.csv`, not the tree** — one row per emitted cmdlet with its module, verb, noun and request path, plus per-module totals in the summary. The 16 hand-written files are the whole reviewable surface: `.gitignore`, `config/ModulesMapping.jsonc`, the two `openApiDocs*` re-slices, six scripts, two
csproj templates, `Program.cs`, and two docs.


peombwa:
The [GitOps/AdvancedSecurity](https://github.com/microsoftgraph/msgraph-sdk-powershell/pull/3733/checks?check_run_id=98251881383) check is failing because `Microsoft.Kiota.Bundle 2.0.0` transitively pulls in `System.Text.Json` v8.0.0, which is affected by two high-severity Denial of Service vulnerabilities:

- **CVE-2024-30105** – .NET Denial of Service Vulnerability
- **CVE-2024-43485** – .NET Denial of Service Vulnerability

To fix this, add an explicit `PackageReference` pinning `System.Text.Json` to `10.0.3` in each affected wrapper `Client.csproj`, the same way it's already pinned in [`Authentication.Core.csproj` (line 30): 
```xml
<PackageReference Include="System.Text.Json" Version="10.0.3" />
```
https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/c7a6aee046575df89e0c891702388aa74c614e01/src/Authentication/Authentication.Core/Microsoft.Graph.Authentication.Core.csproj#L30

Joywambui-maina:
Done in f7b83ce9eb - thanks, and 10.0.3 was the right call over what I was about to use.

Pinned `System.Text.Json` to `10.0.3` in `tools/Templates/WrapperClient.csproj.template` and in all 38 committed `Client.csproj`, matching the pin in `Authentication.Core.csproj` so the repository keeps one answer rather than two.

I had been about to pin 8.0.5, which is the minimum version that patches both advisories (CVE-2024-30105 is fixed in 8.0.4, CVE-2024-43485 in 8.0.5, so 8.0.4 alone is still vulnerable). That would have worked but diverged from the existing convention for no reason.

Verified rather than assumed, since the client projects are netstandard2.0:
- a netstandard2.0 restore with `Microsoft.Kiota.Bundle 2.0.0` plus the pin resolves `System.Text.Json/10.0.3`; without it, `8.0.0`
- `Authentication.Core` targets `netstandard2.0;net6.0;net472`, so 10.0.3 was already proven on that framework here
- Mail rebuilds clean with the pin

Diff is 39 files, +117/-0 - the template and the csprojs only, no generated output.

config/ModulesMapping.jsonc
```diff
@@ -16,7 +16,7 @@
   "DeviceManagement.Administration": "^deviceManagement.(virtualEndpoint.*|.*Partner.*|.*Certificate.*|.*role.*|deviceManagement(DomainJoinConnector|ExchangeConnector|ExchangeOnPremisesPolicy)|groupPolicy(Category|Definition|DefinitionFile|MigrationReport|ObjectFile|UploadedDefinitionFile)|auditEvent|cartToClassAssociation|comanagementEligibleDevice|deviceAndAppManagementRoleAssignment|intuneBrandingProfile|iosUpdateDeviceStatus|mobileThreatDefenseConnector|ndesConnector|resourceOperation|restrictedAppsViolation|termsAndConditions)",
   "DeviceManagement.Enrollment": "^deviceManagement.(.*Enrollment.*|.*Autopilot.*|.*depOnboarding.*|importedDeviceIdentity|onPremisesConditionalAccessSettings|windowsFeatureUpdateProfile)$|^roleManagement.roleManagement$|^roleManagement.rbacApplicationMultiple$|^roleManagement.unifiedRbacApplication$",
   "DeviceManagement.Functions": "^deviceManagement.*.Functions$",
-  "DirectoryObjects": "^directoryObjects\\.|^directory.publicKeyInfrastructureRoot$",
+  "DirectoryObjects": "^directoryObjects\\.",
```
peombwa, line 19: Why are you removing `directory.publicKeyInfrastructureRoot` from `DirectoryObjects` in the mapping file? The tag is still being used in main. This will cause a regression when v3 ships.

peombwa (changes requested):

config/ModulesMapping.jsonc
Joywambui-maina: I may be missing something but I looked at main directly: src/DirectoryObjects has zero files mentioning PublicKeyInfrastructure, while src/Identity.DirectoryManagement has 18 across v1.0 and beta, and all 16 v1.0
rows in MgCommandMetadata.json are Identity.DirectoryManagement. So the tag is claimed in the mapping but doesn't appear to produce output in DirectoryObjects. I removed it because v3 gives Identity.DirectoryManagement "^directory\." so both modules claimed the same surface and generated 8 paths / 16 operations twice.

peombwa: This is an interesting find. So we had a tag in the mapping that was removed via directives in https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/main/src/DirectoryObjects/DirectoryObjects.md. Could you please create an issue so we can ensure that this removal also makes it to v2 as well?

peombwa (approved):

config/ModulesMapping.jsonc
Joywambui-maina: Created issue #3752 to address this.

PR #3746, opened 2026-08-25
feat(wrapper-generator): emit -All and nextLink following for list cmdlets

Joywambui-maina:
Changes proposed:

-Emit an -All switch on every list-shaped cmdlet that follows each non-empty @odata.nextLink to exhaustion, streaming each page to the pipeline before the next request goes out
-Declare the switch on the public dispatcher too, so it reaches the worker — the binder rejects an undeclared parameter before forwarding ever happens
-Cap the total with a bound -Top at whole-page granularity, matching the published ListCmdlet's shipped semantics
-Warn once when a first response still carries a nextLink and -All wasn't supplied — no extra request. This is deliberately stronger than the published SDK, which truncates silently; approved in the design spec
-Continue through kiota's WithUrl, re-applying headers only, so ConsistencyLevel and caller -Headers survive page 2 onward while the link carries its own query state
-Let a pipeline stop (Select-Object -First N, Ctrl+C) pass through the shared catch instead of being re-branded a Graph failure
Record the decisions and their evidence in tools/WrapperGenerator/docs/pagination.md


tools/WrapperGenerator/CmdletEmitter.cs
```diff
@@ -715,13 +737,39 @@ protected override void ProcessRecord()
...
+                {{{fetchedInit}}
+                    var nextLink = result?.OdataNextLink;
+                    while (!string.IsNullOrEmpty(nextLink) && !Stopping{{capCondition}})
+                    {
+                        // The nextLink already carries the original query state, and a raw-URL
+                        // builder ignores templated query parameters anyway - so the continuation
+                        // re-applies headers only; query bindings here would be dead code.
+                        result = client.{{naming.BuilderExpression}}.WithUrl(nextLink).GetAsync(requestConfiguration =>
+                        {{{continuationHeaders}}
+                        }).GetAwaiter().GetResult();
```
peombwa: No cancellation token forwarded to GetAsync. The !Stopping check only fires between pages. A long-running single-page fetch (large page, slow network) won't respond to Ctrl+C until it completes. The doc acknowledges this as a known limitation, but consider whether threading this.StoppingToken (if available in the base class) through to GetAsync is feasible.

peombwa (approved):

tools/WrapperGenerator/CmdletEmitter.cs
Joywambui-maina: Done in 8ecfb88585 - and your hedge was the right instinct, `PipelineStopToken` turned out to be a trap.

It *is* declared by PowerShellStandard.Library 5.1.1 (as a `CancellationToken`), so it compiles. But it does not exist in the Windows PowerShell 5.1 runtime, which these netstandard2.0 modules also target - I checked all three:

| | PipelineStopToken |
|---|---|
| PowerShellStandard.Library 5.1.1 (compile) | present |
| pwsh 7.6.5 | present |
| Windows PowerShell 5.1 | **absent** |

So it would have compiled, passed every gate on PowerShell 7, and thrown for 5.1 users. `StopProcessing` is virtual on all three, so the token is raised from there instead: a `CancellationTokenSource` on `GraphClientCmdlet`, cancelled in `StopProcessing`, exposed as `StoppingToken`, and handed to the paging `GetAsync`.

One thing that fell out of it: the emitted catch was `when (ex is not PipelineStoppedException)`, which would have swallowed the resulting `OperationCanceledException` and surfaced Ctrl+C as `GraphRequestFailed`. It now excludes cancellation too, so a stop ends the pipeline instead of reporting a fake request failure.

Scoped to the emitter and runtime only - the committed corpus regenerates on #3743, which owns that output, so this PR stays generator-only.

peombwa (dismissed):

peombwa (approved):

PR #3753, opened 2026-08-27
build(wrapper): add the internal-feed release pipeline and install guide

Joywambui-maina:
### Changes proposed in this pull request

- Add `.azure-pipelines/wrapper-release.yml`: builds and packs the wrapper modules from committed sources, verifies packages were actually produced before signing, ESRP-signs, and publishes to the project-scoped `Graph Developer Experiences/MSGraph_PowerShell_V3_Build` feed. Mirrors `sdk-release.yml` - same 1ES template, security templates (pre-checks, guardian-analyzer, codesign-nuget, post-checks) and `1ES.PublishNuget@1` task.
- Manual trigger only, and `Publish` defaults to `false`, so the first runs produce a signed `drop` artifact without pushing packages to the feed.
- Version and prerelease flow through `Build-WrapperModule.ps1 -ModuleVersion/-Prerelease` (`3.0.0-alpha<build id>`); nothing edits `ModuleMetadata.json`, which belongs to the v2 release train.
- Add `docs/install-wrapper-modules-from-feed.md`: feed registration and `Install-Module -AllowPrerelease` instructions for testers.

Note: a run can only succeed once the wrapper stack through #3740 has merged (the script's `-ModuleVersion`/`-Prerelease` surface and the package identity fixes land there). The pipeline is inert until then - manual trigger, no CI hooks.

### Other links

- #3740, #3712

Issue #3712, opened 2026-08-10
WrapperGenerator: package generated modules as installable nupkgs (per-module nuspec)


Joywambui-maina:
Summary
Generated modules can be built and imported locally, but there's no way to hand one to another person. Emit a nuspec per generated module and pack it into a .nupkg, so testers can install modules from a feed and give feedback.

What's working now
Build-WrapperModule.ps1 builds any module into an importable PSD1 under artifacts/, and all 35 cmdlet-producing v1.0 modules build and import. The repo already has a packaging convention to follow: every shipped module has a nuspec next to its manifest, packed by PackModule.ps1 / NuspecHelper.ps1.

How can we make it better
Following the existing convention, the generator (or the packaging script) should emit a nuspec per module and pack a .nupkg, versioned as a prerelease. Two decisions to make along the way: what package IDs to use (the real Microsoft.Graph.* names collide with the published SDK on PSGallery, so previews need either a different ID or a separate feed), and where testers install from (GitHub Packages feed, release artifacts, or a private feed).

Joywambui-maina:
Status: the packaging mechanics are done and verified; what's left on this issue are the distribution decisions in #3740 

Done (on feat/wrapper-packaging, stacked on the runtime PR):

Every cmdlet-producing v1.0 module — 38 now, since operation shapes made three action/function-only modules produce cmdlets — packs into a per-module nupkg, versioned as a prerelease (2.39.0-wrapperpreview01).
The nuspec declares Microsoft.Graph.Authentication as a dependency, so install auto-resolves it from the same repository. Proven three rounds from a clean layout: dependency pulled, module imported, live Graph call succeeded with nothing else on the module path.
Module GUIDs are now stable across builds (derived, no lookup table), so update identity survives handout iterations.
All 38 packages verified three rounds each: dependency, label, distinct stable GUIDs.

Waiting for confirmation before the packages go out:

Package IDs — bless/veto Microsoft.Graph.Wrapper.<Module>.
Feed — internal PowerShell_V2_Build push vs direct nupkg sharing (the folder-install path is proven either way).
Host support — pwsh 7+ only, or 5.1 required (extra signing/matched-set work if so).
Prerelease label — wrapperpreview01 unless told otherwise; one-line change.

Joywambui-maina:
Confirmed. 
[...]

PR #3756, opened 2026-08-27
build(wrapper): add the internal-feed release pipeline and install guide (main)

Joywambui-maina:

- Add `.azure-pipelines/wrapper-release.yml`: builds and packs the wrapper modules from committed sources, verifies packages were actually produced before signing, ESRP-signs, and publishes to the project-scoped `Graph Developer Experiences/MSGraph_PowerShell_V3_Build` feed. Mirrors `sdk-release.yml` - same 1ES template, security templates (pre-checks, guardian-analyzer, codesign-nuget, post-checks) and `1ES.PublishNuget@1` task.
- Manual trigger only, and `Publish` defaults to `false`, so the first runs produce a signed `drop` artifact without pushing packages to the feed.
- Version and prerelease flow through `Build-WrapperModule.ps1 -ModuleVersion/-Prerelease` (`3.0.0-alpha<build id>`); nothing edits `ModuleMetadata.json`, which belongs to the v2 release train.
- Add `docs/install-wrapper-modules-from-feed.md`: feed registration and `Install-Module -AllowPrerelease` instructions for testers.

Same content as #3753 (which targets `powershell-v3`); this twin lands it on `main` so the AzDo pipeline definition can register against the default branch. On `main` the pipeline is inert - the wrapper tooling it invokes lives on the `powershell-v3` line, so runs should select that branch until v3 ships.

### Other links

- #3753, #3740, #3712


docs/install-wrapper-modules-from-feed.md
peombwa, line 1: No need to include the markdown for main. We are only interested in the yml pipeline for main.

Joywambui-maina: Removed in c6010c7f3b - the guide stays on #3753 for powershell-v3, where the wrapper modules live; this PR now carries only the pipeline YAML.

peombwa (approved):

